PR #2, opened 2023-09-05
 Add colors to passed and failed messages

Tito-Kati:
Example:

imagen

If it's okay I would also remove the emojis, since they don't have good support on all operating systems/fonts.

Personally, the terminals have many formatting options; take a look at how [Jest.js](https://jestjs.io) uses them.


Chemaclass (approved):
Wow, I love this! Thanks

PR #4, opened 2023-09-06
chore: created assertNotContains not using the other assertion

khru:
I've added the `assertNotContains` funtion to the framwork, not using the `assertContains` to don't make any test fail for dependency between functions


Tito-Kati (approved):

Chemaclass (approved):
Thanks!

PR #7, opened 2023-09-06
Avoid stop running tests after a failing one test

Chemaclass:
## 📚 Description

Currently, after a test fails, then the `test_runner` stops, and no more tests are executed.

## 🔖 Changes

Move the `exit` from failing test to the end of the `test_runner`, so it will run all tests and keep in memory the number of failing/passing tests to display as an output later.


Tito-Kati (approved):

fabriziofs (approved):

JesusValera (approved):
Great 😃👍🏼👍🏼

PR #15, opened 2023-09-07
Allow ignoring a test using a comment skip

Chemaclass:
## 📚 Description

Allow skip/ignore an actual test if inside the test function there is a comment like `# skip`

```bash
function test_ignore_failing_test() {
  # skip
  assertEquals "1" "2"
}
```

Screenshot 2023-09-07 at 22 23 13

### DIFF
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -2,6 +2,7 @@
 
 export TEST=true
 
+export normalizeFnName
 export assertEquals
 export assertContains
 export assertNotContains
--- a/src/test_runner.sh
+++ b/src/test_runner.sh
@@ -1,16 +1,26 @@
 #!/bin/bash
 
 # shellcheck disable=SC2155
-# shellcheck disable=SC2034
+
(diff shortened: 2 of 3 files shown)


Tito-Kati:
It's better not to add all this complexity for now. To skip a test right now it's as simple as adding skip_ in front of the test function name. This also opens the door to using todo_ or ignore_.

Example:
```
function skip_test_successful_assertNotContains() {
    ...
}
```

khru:
In addition to this: 
> It's better not to add all this complexity for now. To skip a test right now it's as simple as adding skip_ in front of the test function name. This also opens the door to using todo_ or ignore_.
> 
> Example:
> 
> ```
> function skip_test_successful_assertNotContains() {
>     ...
> }
> ```

I would also be careful with using comming because we could face things like:

```
function test_ignore_failing_test() {
  # skip the number 2
  assertEquals "1" "3"
}
```

PR #27, opened 2023-09-08
chore: added assert matches, refactor global variables to start with underscored and add watch to the Makefile

khru:
Related to: https://github.com/Chemaclass/bashunit/issues/23

Allow asserting by using regular expressions:

```bash
assertMatches ".*Pinux*" "GNU/Linux"
```

### DIFF
--- a/src/console_results.sh
+++ b/src/console_results.sh
@@ -19,5 +19,5 @@ ${COLOR_FAINT}Total assertions:${COLOR_DEFAULT} ${COLOR_BOLD}${totalAssertions}$
   fi
 }
 
-# Set a trap to call render_result when the script exits
+# Set a trap to call renderResult when the script exits
 trap 'renderResult $_TOTAL_TESTS $_TOTAL_ASSERTIONS_PASSED $_TOTAL_ASSERTIONS_FAILED' EXIT
--- a/src/console_results.sh
+++ b/src/console_results.sh
@@ -1,0 +1,23 @@
+export renderResult
+
+function renderResult() {
+  local totalTests=$1
+  local totalPassed=$2
+  local totalFailed=$3
+
+  echo ""
+  local totalAssertions=$((totalPassed + totalFailed))
+  printf "\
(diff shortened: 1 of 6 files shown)


Tito-Kati (approved):
test/watch 👌🏻 thank you mate

Chemaclass (approved):
Nice! Thank you!

Is there a convention in bash to start globals with underscore, @khru? (eg: `_TOTAL_TESTS`)

src/console_results.sh
```diff
@@ -0,0 +1,23 @@
...
+
+  if [ "$totalFailed" -gt 0 ]; then
+    printf "${COLOR_FAINT}Total assertions failed:${COLOR_DEFAULT} ${COLOR_BOLD}${COLOR_FAILED}${totalFailed}${COLOR_DEFAULT}\n"
+    exit 1
+  else
+    printf "${COLOR_ALL_PASSED}All assertions passed.${COLOR_DEFAULT}\n"
+  fi
+}
+
+# Set a trap to call render_result when the script exits
```
Chemaclass, line 22:
```suggestion
# Set a trap to call renderResult when the script exits
```

khru:
No there is not as far as I am aware of, but as part of a framework we should start using namespaces and names on variables that cannot collide with other variables that the bash scripts we are testing use

> Nice! Thank you!
> 
> Is there a convention in bash to start globals with underscore, @khru? (eg: `_TOTAL_TESTS`)


Example:

Let's say that you have in your code a variable with the name `TOTAL_TESTS` Whenever we run our test we will override that variable to ours, and I think that's not the ideal scenario

src/console_results.sh
khru: Related to this, the convention on Bash it's to use for the variables always Upper snake case, so my suggestion it's to add a https://github.com/Chemaclass/bashunit/issues/20 static analysis tool to the local repo and prepare a pre-commit for it.

A conversation about this is: https://unix.stackexchange.com/questions/42847/are-there-naming-conventions-for-variables-in-shell-scripts#answer-42849

Although for example, this is not in these guidelines https://google.github.io/styleguide/ the community uses this

Chemaclass (approved):

PR #30, opened 2023-09-08
chore: add shellcheck

khru:
# Linux
- [X] Check that the bin works
- [X] Check that the Makefile works
# Mac
- [x] Check that the bin works
- [x] Check that the Makefile works

Related to: https://github.com/Chemaclass/bashunit/issues/20

Source of the static tool: https://github.com/koalaman/shellcheck#from-your-terminal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,34 @@
 SHELL=/bin/bash
+
+OS:=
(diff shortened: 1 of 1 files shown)


Tito-Kati:
imagen
In my shell emojis looks weird as usual 😄

Issue #26, opened 2023-09-08
Listing not test but assertions


khru:
On the list of tests we are executing we are not showing the list of test running but the list of assertions happening:

![image](https://github.com/Chemaclass/bashunit/assets/6353105/f89682ef-a798-459d-9586-1c3bf0224973)

This bug could relate to this Issue: https://github.com/Chemaclass/bashunit/issues/18

Tito-Kati:
Currently, all assertions are printing both their 'passed' and 'failed' messages, when in reality, assertions should only print the 'failed' messages. For a test to be considered 'passed', all assertions within that test must pass.

Having said that, our assertions can currently accept an optional third argument, which defines the text displayed when the assertion passes. Ideally, this text should always be the name of the test, and this third argument message should only be displayed when the assertion fails.

As examples, I mention [PHPUnit's implementations](https://docs.phpunit.de/en/10.3/assertions.html#boolean) (which support it natively, see $message) and Jest's, which supports it [through a plugin](https://github.com/mattphillips/jest-expect-message).

By the way, I don't see the implementation of this feature as mandatory, but I do find it interesting and practical.

PR #40, opened 2023-09-09
Listing tests

Tito-Kati:
## 📚 Description

I've developed the output format of the renderResult function to align with what was discussed in issue #26.

Now it counts tests and assertions separately. A test is considered failed when any assertion fails or when a non-zero code is returned within the test.

A test is counted as passed when all assertions inside the test pass.

Tests and assertions that have passed or failed are not displayed if there aren't any, similar to Jest's style

And many variable and function names have been refactored because @Chemaclass and I got carried away.

It's not Friday, so it can be deployed to production without any issues 😉.

output example


Chemaclass (approved):

JesusValera (approved):
Awesome stuff 😎

khru:
@Tito-Kati Question. Does this solves the issue with the parallel run?

Tito-Kati:
> @Tito-Kati Question. Does this solves the issue with the parallel run?

No 😞

PR #53, opened 2023-09-11
Add README.md inside docs/

Chemaclass:
## 📚 Description

Create a `README.md` to display all assert functions when you go to the `docs/` folder

Screenshot 2023-09-11 at 14 30 57

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,25 @@
+
+<p align="center">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="../docs/assets/logo_name_dark.svg">
+    <img alt="bashunit" src="../docs/assets/logo_name.svg" width="400">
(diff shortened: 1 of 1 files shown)


JesusValera (commented):
As an idea is ok, I like it, but in the case of adding some description next to the `assertXXX`, we could add something more meaningful 🙂

khru:
@Chemaclass I've added to the ``functional/logic_test.sh` all this test

image

Should I put them into the example or we should update the module with the example inside?

khru:
My suggestion for changes would be this: https://github.com/Chemaclass/bashunit/pull/55

Chemaclass:
> As an idea is ok, I like it, but in the case of adding some description next to the `assertXXX`, we could add something more meaningful 🙂

Feel free to add a description in a follow up commit

PR #55, opened 2023-09-11
Change suggestions

khru:


Chemaclass (approved):

tests/functional/logic_test.sh
```diff
@@ -4,6 +4,50 @@ ROOT_DIR="$(dirname "${BASH_SOURCE[0]}")"
 
 SCRIPT="$ROOT_DIR/logic.sh"
 
-function test_your_logic() {
-  assertEquals "expected 123" "$($SCRIPT "123")" "Functional test working!"
```
Chemaclass, line 8: I would keep here only the `test_your_logic` as how it was before, because you already moved the tests to the example dir

Chemaclass:
@khru, I get this on my local when running the linter `make lint`. What command are you running?
Screenshot 2023-09-11 at 16 15 23

tests/functional/logic_test.sh
Chemaclass: What do you think @khru @Tito-Kati ?

khru:
> Screenshot 2023-09-11 at 16 28 08

Now after the update:

image

PR #58, opened 2023-09-11
Throw error on duplicated tests

Chemaclass:
## 📚 Description

Issue: https://github.com/Chemaclass/bashunit/issues/56

If you have to test with the same name, we should throw an error or say something, because it could be a bad situation

## 🔖 Changes

- Refactor extract function `getFunctionsToRun()`
- assertGeneralError when duplicate functions


JesusValera (approved):

Tito-Kati (approved):
I love it

PR #70, opened 2023-09-13
Update README.md

CosmeValera:
## 🔖 Changes

- Include Java, JavaScript and Python examples. Remove composer / php example. 
- "Check out" sounds more natural than "see".

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,11 @@ function test_your_logic() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Tito-Kati (approved):

Chemaclass (approved):
It's great to add those examples, but I would keep the PHP one as well

khru (approved):

PR #75, opened 2023-09-13
Add editorconfig lint

khru:


Tito-Kati (approved):
Oh yes, I really like this. When do we merge it?

PR #90, opened 2023-09-15
Add bpkg.json

Chemaclass:
## 📚 Description

Related to: https://github.com/bpkg/bpkg.github.io/pull/82 

## 🔖 Changes

- Create a `bpkg.json` following https://bpkg.sh/guidelines/

### DIFF
--- a/bpkg.json
+++ b/bpkg.json
@@ -1,0 +1,7 @@
+{
+  "name": "bashunit",
+  "description": "A simple testing library for bash scripts.",
+  "scripts": [ "bashunit" ],
+  "install": "make install",
+  "global": "true"
+}
--- a/bpkg.json
+++ b/bpkg.json
@@ -2,6 +2,6 @@
   "name": "bashunit",
   "description": "A simple testing library for bash scripts.",
   "scripts": [ "bashunit" ],
-  "install": "make install",
-  "global": "true"
+  "install": "install -b bashunit ${PREFIX:-/usr/local}/bin/bashunit",
+  "global": "false"
 }


bpkg.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "bashunit",
+  "description": "A simple testing library for bash scripts.",
+  "scripts": [ "bashunit" ],
+  "install": "make install",
```
khru, line 5: Be careful with this, make does not work on windows

Chemaclass: Actually, I was wondering about this "install" key. Is it really [necessary](https://bpkg.sh/guidelines/#install)? What should we put there?  🤔

Chemaclass: Actually, I was wondering about this "install" key. Is it really [necessary](https://bpkg.sh/guidelines/#install)?
What should we put there, @samlikins?  🤔

khru: I'm guessing that it's the script that will install our software, but I'm not sure

samlikins: The `scripts` property is an array of paths (to files in the project) which you want installed in the `deps` directory, when your project is included as a dependency.  When installing globally, the `install` property is a command string for performing a system install.  The example uses a `Makefile`, but it could be a shell script in your project.  Take a look at some of the other projects.  The `bash-util-lib` project uses an `install` command.  The `dots` project uses a shell script.  The `JSON` project uses a simple `cp` command.  Using the `PREFIX` environment variable will provide a system `bin` parent directory, if performing an install as `root`, a user directory otherwise.

samlikins:
Once you get this file into your project, you can start testing out your project with the BPKG utility, even before listing it on bpkg.sh.

bpkg.json
samlikins: The `global` property should be removed, if you want it to be installed in projects, instead of system wide.  Try the following for the install property: `install -b bashunit.sh ${PREFIX:-/usr/local}/bin/bashunit`, this will be run when someone wants to install as a system executable.

Chemaclass: Great. I will try that for now. Thank you!

khru (commented):
@Chemaclass we would need documentation and reference for new contributors on:
- How it works?
- What it is?
- How to test it?

khru:
> Once you get this file into your project, you can start testing out your project with the BPKG utility, even before listing it on bpkg.sh.

Merge, pleas let us know how it goes

samlikins:
@khru, did you need me to perform some action?  My suggestion to test `bashunit` once the `bpkg.json` is in the project, was for someone in your team to try your project out.  From system wide install as root and non-root user to as dependency in another project (even just a test project). Document using BPKG for installing or using as dependency.

Issue #56, opened 2023-09-11
Throw error on test name duplicated


khru:
If you have to test with the same name, we should throw an error or say something, because it could be a bad situation

Tito-Kati:
Related to #55

PR #101, opened 2023-09-17
Stop runner when found duplicate functions

Chemaclass:
## 📚 Description

Issue: https://github.com/TypedDevs/bashunit/issues/56 

Stop the execution of the test runner if you have two tests with the same name.

## 🔖 Changes

- Terminate the runner if duplicated tests were found

## 🖼️ Screenshots

Example:

Screenshot 2023-09-19 at 12 36 06

## 🏗️ TODO

This current implementation works only when using the syntax `function foo () {...`, but in bash you can define functions without `function` and without the parentheses, so these two forms should also be valid and implemented in a follow up PR
- `yourFunc() { ...`
- `function yourFunc { ... `


tests/unit/assert_test.sh
```diff
@@ -81,91 +81,91 @@ function test_unsuccessful_assertNotMatches() {
 }
 
 function test_successful_assertExitCode() {
-  function fake_function() {
+  function fake_function1() {
```
khru, line 84: why did we change the name to the functions on each test?

src/runner.sh
```diff
@@ -99,12 +98,24 @@ function Runner::loadTestFiles() {
...
+    duplicates="$(Helper::getDuplicateFunctions "$test_file")"
+    if [[ -n "$duplicates" ]]; then
+      _DUPLICATED_FOUND=true
+      echo "Duplicate functions found in '$test_file'"
+      IFS=',' read -ra elements <<< "$duplicates"
+      for element in "${elements[@]}"; do
+          echo "- $element"
+      done
+      exit 1
+    fi
```
khru, line 117: Why do we have a function that returns a string concatenated with the functions and not just whether it is duplicated or not, I have the perception that this logic should be somewhere else and that this is leaking the implementation details into a file that should not have this responsibility

khru (changes requested):
Why if there is duplicated the exit code it's 0 at the end of the execution of bashunit.

test to prove it:
```bash
function test_with_duplicates_should_end_with_error_code() {
  assertGeneralError "$(./bashunit "$(dirname "${BASH_SOURCE[0]}")/fixtures/duplicate_functions.sh")"
}
```

image

.editorconfig
```diff
@@ -9,7 +9,7 @@ trim_trailing_whitespace = true
 [**.sh]
 indent_style = space
 indent_size = 2
-max_line_length = 120
+max_line_length = 122
```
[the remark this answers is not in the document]
Tito-Kati: https://github.com/TypedDevs/bashunit/pull/82/files#diff-14ed4071bf97ddd640c335f1d4768f2aeda369f75b3dcd351cf3890bea8a9c90R115-R119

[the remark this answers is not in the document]
khru: but it should not be needed it anymore right?

tests/unit/assert_test.sh
Tito-Kati: This might be due to a planning oversight. We want to find the duplicated test functions, not all functions. I believe adding to the regular expression that filters the functions to only search for those starting with test_ would be sufficient.

Chemaclass: Great idea 💡  I will do that

src/helpers.sh
```diff
@@ -21,12 +21,28 @@ function Helper::normalizeTestFunctionName() {
   echo "$result"
 }
 
+#
+# @param $1 string Eg: "your_script.sh"
+#
+# @return string Eg: "func1,func2"
+#
+function Helper::getDuplicateFunctions() {
```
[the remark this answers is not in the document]
Chemaclass: I changed it to `checkDuplicateFunctions` and do an early `exit 1` in case of duplicates.

src/runner.sh
Chemaclass: I changed this to something like this:
```bash
local duplicates="$(Helper::checkDuplicateFunctions "$test_file")"
if [[ $duplicates == true ]]; then
  State::setDuplicatedFound
  echo "Duplicate functions found in '$test_file'"
  exit 1
fi
```

src/helpers.sh
```diff
@@ -21,12 +21,28 @@ function Helper::normalizeTestFunctionName() {
...
+  grep -E '^\s*(function)?\s*[a-zA-Z_][a-zA-Z_0-9]*\s*\(\)?\s*{' "$script" | \
+    awk '{gsub(/\(|\)/, ""); print $2}' | \
+    sort | \
+    uniq -d | \
+    tr '\n' ','| \
+    sed 's/,$//'
+}
+
 #
 # @param $1 string Eg: "prefix"
```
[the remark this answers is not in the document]
Chemaclass: Thanks. I will do this in a follow up PR

src/console_results.sh
```diff
@@ -3,12 +3,19 @@
 _START_TIME=$(date +%s%N);
 
 function Console::renderResult() {
+  local exit_code=$?
+  if [[ "$exit_code" == "$_DUPLICATED_TEST_FUNCTIONS_FOUND_ERROR_CODE" ]]; then
+    echo "Duplicate functions found"
```
Tito-Kati, line 8: Should use text with a red background? (\e[41m)

src/helpers.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
+_DUPLICATED_TEST_FUNCTIONS_FOUND_ERROR_CODE=100
```
Tito-Kati, line 3: From which number were the codes free, could it have been from 128? @khru ?

src/runner.sh
```diff
@@ -110,10 +114,30 @@ while [[ $# -gt 0 ]]; do
...
-Runner::loadTestFiles
 
-trap 'Console::renderResult '\
-'"$(State::getTestsPassed)" '\
-'"$(State::getTestsFailed)" '\
-'"$(State::getAssertionsPassed)" '\
-'"$(State::getAssertionsFailed)"' EXIT
+function Runner::loadTestFiles() {
+  if [ ${#_FILES[@]} -eq 0 ]; then
+    echo "Error: At least one file path is required."
```
Tito-Kati, line 120: Should use text with a red background here too?

tests/unit/helpers_test.sh
```diff
@@ -61,6 +69,10 @@ function test_no_function_is_executed_with_execute_function_if_exists() {
 function test_unsuccessful_unsetIfExists() {
   assertGeneralError "$(Helper::unsetIfExists "fake_function")"
 }
+function test_unsuccessful_unsetIfExists() {
+  # TODO: this is duplicated and it should failed; should be fixed on 'feat/stop-exec-when-found-duplicate'
```
Tito-Kati, line 73: ?

src/runner.sh
Chemaclass: Good idea, but I would suggest doing this in a follow up PR

tests/unit/helpers_test.sh
Chemaclass: This was an intentional duplicated test to force the error of duplicated test functions. I forgot to remove it, thanks

src/helpers.sh
Chemaclass: I couldn't make it using this approach of exit codes. I changed it to use a global flag `_DUPLICATED_TEST_FUNCTIONS_FOUND`

Chemaclass:
> Why if there is duplicated the exit code it's 0 at the end of the execution of bashunit.
> 
> test to prove it:
> 
> ```shell
> function test_with_duplicates_should_end_with_error_code() {
>   assertGeneralError "$(./bashunit "$(dirname "${BASH_SOURCE[0]}")/fixtures/duplicate_functions.sh")"
> }
> ```
> 
> image

I've tried different approaches, but I couldn’t make this happen... Maybe you could try this out. What about working on this idea in a follow up PR to unblock the changes in this one?

src/console_results.sh
```diff
@@ -3,12 +3,18 @@
 _START_TIME=$(date +%s%N);
 
 function Console::renderResult() {
+  if [[ "$(State::isDuplicatedTestFunctionsFound)" == true ]]; then
+    printf "\e[31m> Duplicate test functions found\e[0m\n"
```
Tito-Kati, line 7: Use color constants instead 🦆

tests/unit/fixtures/duplicate_functions.sh
```diff
@@ -0,0 +1,30 @@
...
+function func1() {
+    echo "Function 1"
+}
+
+function func2() {
+    echo "Function 2"
+}
+
+function func1() {
+    echo "Function 1 Duplicate but not test"
```
Tito-Kati, line 13: Shouldn't this duplicated function be in the no_duplicate_functions.sh file? Otherwise, you're not checking if the functions that aren't tests can indeed be duplicated without causing errors.

Tito-Kati (changes requested):

Tito-Kati (approved):

src/runner.sh
Chemaclass: Ok, I will do this in this PR, @Tito-Kati 😝

PR #104, opened 2023-09-19
Refactor loadTestFiles to receive args instead of using globals

Chemaclass:
## 🔖 Changes

- Avoid using globals from `Runner::loadTestFiles`, instead use local arguments.


Tito-Kati (approved):
Nice 🙆🏻‍♂️

JesusValera (approved):

PR #105, opened 2023-09-19
Catch errors when test execution fails

cmayo:
## 📚 Description

In case a test fails because an error inside, it will be marked as failed and it will show the result code error:

imagen

imagen

## ✅ To-do list

- [ ] Make sure that all the pipeline passes

### DIFF
--- a/src/runner.sh
+++ b/src/runner.sh
@@ -28,10 +28,11 @@ function Runner::callTestFunctions() {
 function Runner::runTest() {
   local function_name="$1"
   local current_assertions_failed
+  local test_result_code=0
   current_assertions_failed="$(State::getAssertionsFailed)"
 
   Runner::runSetUp
-  "$function_name"
+  "$function_name" || test_result_code=$?
   Runner::runTearDown
 
   if [[ "$current_assertions_failed" != "$(State::getAssertionsFailed)" ]]; then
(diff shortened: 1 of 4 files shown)


src/runner.sh
```diff
@@ -28,12 +28,19 @@ function Runner::callTestFunctions() {
 function Runner::runTest() {
   local function_name="$1"
   local current_assertions_failed
+  local test_result_code=1
```
Chemaclass, line 31: What about using 0 as the default value, as 0 means success in bash?

Chemaclass (commented):
Awesome, thanks! Could you add an automated test for this new behavior?

src/runner.sh
cmayo: Yes, you are right 👍

khru (changes requested):
This does not solve the full issue:

Example:
```bash
function test_xxx() {
  source "a_non_existing_file"
  assertEquals "true" "true"
}
```
Output
image

cmayo:
The best solution here would be to run the test in a subshell and get the result code like this:

```
(
  set -e
  Runner::runSetUp
  "$function_name"
  Runner::runTearDown
)
local test_result_code=$?
```

So it will stop on the first error inside the test and not wait for the return code like I did in this PR.

The problem with this is that in the subshell we don't have access to the same state variables of the parent shell so we can't increase them from the subshell.

khru:
Exactly, that was what we thought, but we had no luck the last time we tried to implement it in fact that's the next thing we have as one of the things we want to solve.

> The best solution here would be to run the test in a subshell and get the result code like this:
> 
> ```
> (
>   set -e
>   Runner::runSetUp
>   "$function_name"
>   Runner::runTearDown
> )
> local test_result_code=$?
> ```
> 
> So it will stop on the first error inside the test and not wait for the return code like I did in this PR.
> 
> The problem with this is that in the subshell we don't have access to the same state variables of the parent shell so we can't increase the assertion count from the subshell.

cmayo:
I will close this PR, I have created a new one here https://github.com/TypedDevs/bashunit/pull/109

PR #109, opened 2023-09-20
Manage error when test execution fails

cmayo:
## 📚 Description

I have created a new PR (coming from https://github.com/TypedDevs/bashunit/pull/105) from a new branch in this repository so now we can work together on it.

In case a test fails because an error inside, it will be marked as failed and it will show the result code error:

imagen

imagen

The main objective is execute the tests in a subshell with the `set -e`. The problem with this is that in the subshell we don't have access to the same state variables of the parent shell so we can't increase the assertion count from the subshell.

To avoid this problem, the idea is return (printing) the total of assertions failed and passed in the subshell and then in the main runner shell we have to parse this output and get the assertions failed and passed so we can increase them now in the main shell. 

There are some side effects I have changed too:
- Assert and helper functions should return 0 to prevent being stopped
- Test files for testing errors (like tests/acceptance/bashunit_test.sh) need to be executed in a subhell with `set +e` to prevent being catch by the main shell.

### DIFF
--- a/src/state.sh
+++ b/src/state.sh
@@ -12,6 +12,7 @@ function State::getTestsPassed() {
 
 function State::addTestsPassed() {
   ((_TESTS_PASSED++))
+  return 0
 }
 
 function State::getTestsFailed() {
@@ -20,6 +21,7 @@ function State::getTestsFailed() {
 
 function State::addTestsFailed() {
   ((_TESTS_FAILED++))
+  return 0
 }
 
 function State::getAssertionsPassed() {
@@ -27,7 +29,8 @@ function State::getAssertionsPassed() {
 }
 
 function State::addAssertionsPassed() {
-  ((_ASSERTIONS_PASSED++))
+  ((_ASSERTIONS_PASSED++)) || true
+  return 0
 }
 
 function State::getAssertionsFailed() {
@@ -35,7 +38,8 @@ function State::getAssertionsFailed() {
 }
 
 function State::addAssertionsFailed() {
-  ((_ASSERTIONS_FAILED++))
+  ((_ASSERTIONS_FAILED++)) || true
+  return 0
 }
 
 function State::isDuplicatedTestFunctionsFound() {
@@ -44,4 +48,14 @@ function State::isDuplicatedTestFunctionsFound() {
 
 function State::setDuplicatedTestFunctionsFound() {
   _DUPLICATED_TEST_FUNCTIONS_FOUND=true
+  return 0
+}
+
+function State::initializeAssertionsCount() {
+    _ASSERTIONS_PASSED=0
+    _ASSERTIONS_FAILED=0
+}
+
+function State::exportAssertionsCount() {
+  echo "##ASSERTIONS_FAILED=$_ASSERTIONS_FAILED##ASSERTIONS_PASSED=$_ASSERTIONS_PASSED##"
 }
--- a/src/state.sh
+++ b/src/state.sh
@@ -11,8 +11,7 @@ function State::getTestsPassed() {
 }
 
 function State::addTestsPassed() {
-  ((_TESTS_PASSED++))
-  return 0
+  ((_TESTS_PASSED++)) || true
 }
 
 function State::getTestsFailed() {
@@ -20,8 +19,7 @@ function State::getTestsFailed() {
 }
 
 function State::addTestsFailed() {
-  ((_TESTS_FAILED++))
-  return 0
+  ((_TESTS_FAILED++)) || true
 }
 
 function State::getAssertionsPassed() {
@@ -30,7 +28,6 @@ function State::getAssertionsPassed() {
 
 function State::addAssertionsPassed() {
   ((_ASSERTIONS_PASSED++)) || true
-  return 0
 }
 
 function State::getAssertionsFailed() {
@@ -39,7 +36,6 @@ function State::getAssertionsFailed() {
 
 function State::addAssertionsFailed() {
   ((_ASSERTIONS_FAILED++)) || true
-  return 0
 }
 
 function State::isDuplicatedTestFunctionsFound() {
@@ -48,7 +44,6 @@ function State::isDuplicatedTestFunctionsFound() {
 
 function State::setDuplicatedTestFunctionsFound() {
   _DUPLICATED_TEST_FUNCTIONS_FOUND=true
-  return 0
 }
 
 function State::initializeAssertionsCount() {
--- a/tests/functional/logic_test.sh
+++ b/tests/functional/logic_test.sh
@@ -39,6 +39,8 @@ function test_should_validate_a_non_ok_exit_code() {
     return 1
   }
 
+  set +e
+
   fake_function
 
   assert_exit_code "1"
--- a/tests/unit/assert_test.sh
+++ b/tests/unit/assert_test.sh
@@ -113,11 +113,7 @@ function test_unsuccessful_return_assert_exit_code() {
     return 1
   }
 
-  (
-    set +e
-    fake_function
-    assert_exit_code "1"
-  )
+  assert_exit_code "1" "$(fake_function)"
 }
 
 function test_successful_assert_successful_code() {
--- a/tests/unit/assert_test.sh
+++ b/tests/unit/assert_test.sh
@@ -113,9 +113,11 @@ function test_unsuccessful_return_assert_exit_code() {
     return 1
   }
 
-  fake_function
-
-  assert_exit_code "1"
+  (
+    set +e
+    fake_function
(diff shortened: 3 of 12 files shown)


tests/functional/logic_test.sh
```diff
@@ -39,6 +39,8 @@ function test_should_validate_a_non_ok_exit_code() {
     return 1
   }
 
+  set +e
```
Chemaclass, line 42: Note (for the record):
- Using `set -e`, the shell will exit immediately if any command exits with a non-zero status.
- Using `set +e`, the script will continue running even if a command fails.

Chemaclass (approved):

src/state.sh
```diff
@@ -20,22 +21,25 @@ function State::getTestsFailed() {
...
 }
 
 function State::getAssertionsPassed() {
   echo "$_ASSERTIONS_PASSED"
 }
 
 function State::addAssertionsPassed() {
-  ((_ASSERTIONS_PASSED++))
+  ((_ASSERTIONS_PASSED++)) || true
+  return 0
```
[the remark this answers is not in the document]
cmayo: You're right, we need this `|| true` for the other lines too. And I've been checking that we don't need the `return 0` in the assert functions, the returned value on those functions are the same that the last function called inside, so in that case the `State::addAssertionsPassed` one, that would be `0`

khru (changes requested):
This test should be failing but it's passing
```
function test_xxx() {
  assertEquals "$(a_function_that_does_not_exists)"
}
```

khru:
@Chemaclass could you give a hand with resolving the conflicts?

Chemaclass:
For sure! @khru, you can solve the conflict by removing the duplicated `function Runner::loadTestFiles()` after merging `main` into `manage_error_tests`:

Screenshot 2023-09-28 at 15 37 07

khru:
> For sure! @khru, you can solve the conflict by removing the duplicated `function Runner::loadTestFiles()` after merging `main` into `manage_error_tests`:
> 
> Screenshot 2023-09-28 at 15 37 07

image

src/state.sh
```diff
@@ -11,31 +11,31 @@ function State::getTestsPassed() {
 }
 
 function State::addTestsPassed() {
-  ((_TESTS_PASSED++))
+  ((_TESTS_PASSED++)) || true
```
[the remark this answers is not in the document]
cmayo: This is because when we are running the script with `set -e` to stop on the failing command, the arithmetic operation is evaluated and it's returning a non zero value. 

https://stackoverflow.com/a/66824557

You can run this example:

```
#!/bin/bash

set -e

test_var=0
(( test_var++)) 
echo "$test_var"
```

The return value on `(( test_var++ ))` is 1 so with the `set -e` enabled it will stop in this point. If you add `(( test_var++)) || echo $?` you can see the returned value

cmayo:
> This test should be failing but it's passing
> 
> ```
> function test_xxx() {
>   assertEquals "$(a_function_that_does_not_exists)"
> }
> ```
> 
> @cmayo If you need any help let us know we could pair if you need any help.
> 
> CC: @Tito-Kati @Chemaclass

The problem here is how to propagate the `set -e` option because it depends also on the Bash version. For example, macOS is still using the version 3 of bash when Linux is using the version 5 :-D

I have found some interesting links:

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
https://mywiki.wooledge.org/BashFAQ/105
https://dougrichardson.us/notes/fail-fast-bash-scripting.html

cmayo:
> > For sure! @khru, you can solve the conflict by removing the duplicated `function Runner::loadTestFiles()` after merging `main` into `manage_error_tests`:
> > Screenshot 2023-09-28 at 15 37 07
> 
> image

@khru @Chemaclass I've rebased this branch and resolve the conflicts. Also made a fix on the spy function related to the arithmetic operation in the eval.

tests/unit/helpers_test.sh
```diff
@@ -67,7 +67,7 @@ function test_no_function_is_executed_with_execute_function_if_exists() {
 }
 
 function test_unsuccessful_unsetIfExists() {
-  assert_general_error "$(Helper::unsetIfExists "fake_function")"
+  assert_successful_code "$(Helper::unsetIfExists "fake_function")"
```
Chemaclass, line 70: Isn't the test checking the unsuccessful behavior?

cmayo: In my opinion the `unsetIfExists` function should not return any error in case the function doesn't exists. As the function name said, it's only unset if it exists.

cmayo:
It's not easy no catch all errors that happens inside a subshell. It's deppends on how the are executed, with `( ), $( ), { }`, ... and also the `set -e` doesn't stop on all errors. This is an interesting link:

https://mywiki.wooledge.org/BashFAQ/105
https://stackoverflow.com/q/22630363

tests/unit/helpers_test.sh
Chemaclass: Oh, good point! I agree

Chemaclass:
@cmayo, if you call a non existing function, the test is not failing (same behaviour as in main). Is this known/intended?

```bash
function test_xxx() {
  assertEquals "$(a_function_that_does_not_exists)"
}
```

cmayo:
> @cmayo, if you call a non existing function, the test is not failing (same behaviour as in main). Is this known/intended?
> 
> ```shell
> function test_xxx() {
>   assertEquals "$(a_function_that_does_not_exists)"
> }
> ```

Yes, I know it. I'm trying to apply one of the solutions applied in this document https://fvue.nl/wiki/Bash:_Error_handling but as you can read on it, sometime is not possible to catch all errors.

Chemaclass:
> > @cmayo, if you call a non existing function, the test is not failing (same behaviour as in main). Is this known/intended?
> 
> Yes, I know it. I'm trying to apply one of the solutions applied in this document https://fvue.nl/wiki/Bash:_Error_handling but as you can read on it, sometime is not possible to catch all errors.

I would vote for merging this PR as it is now, as there are already improvements _especially_ running each test function in a subshell. And working on the further improvements in a follow up PR(s). 
What do you think @cmayo, @khru , @Tito-Kati ?

Chemaclass:
> > @cmayo, if you call a non existing function, the test is not failing (same behaviour as in main). Is this known/intended?
> 
> Yes, I know it. I'm trying to apply one of the solutions applied in this document https://fvue.nl/wiki/Bash:_Error_handling but as you can read on it, sometime is not possible to catch all errors.

I would vote for merging this PR as it is now, as there are already improvements _especially_ running each test function in a subshell. And working on the further improvements in a follow up PR(s). 
What do you think @cmayo, @khru , @Tito-Kati ?

cmayo:
> > > @cmayo, if you call a non existing function, the test is not failing (same behaviour as in main). Is this known/intended?
> > 
> > 
> > Yes, I know it. I'm trying to apply one of the solutions applied in this document https://fvue.nl/wiki/Bash:_Error_handling but as you can read on it, sometime is not possible to catch all errors.
> 
> I would vote for merging this PR as it is now, as there are already improvements _especially_ running each test function in a subshell. And working on the further improvements in a follow up PR(s). What do you think @cmayo, @khru , @Tito-Kati ?

For me it's ok merging it now and work on further improvements.

The main problem with this test:

```
function test_xxx() {
  assertEquals "$(a_function_that_does_not_exists)"
}
```

is that the command substitution not always return the exit code. We should advise that a proper way to write this type of test could be:

```
function test_xxx() {
  local expected
  expected=$(a_function_that_does_not_exists);

  assertEquals "$expected"
} 
```

Writing the test in this way it will fail as expected.

tests/unit/helpers_test.sh
```diff
@@ -67,7 +67,7 @@ function test_no_function_is_executed_with_execute_function_if_exists() {
 }
 
 function test_unsuccessful_unsetIfExists() {
```
Tito-Kati, line 69:
```suggestion
function test_successful_unsetIfExists_non_existent_function() {
```

tests/unit/assert_test.sh
```diff
@@ -113,9 +113,11 @@ function test_unsuccessful_return_assert_exit_code() {
...
   }
 
-  fake_function
-
-  assert_exit_code "1"
+  (
+    set +e
+    fake_function
+    assert_exit_code "1"
+  )
```
Tito-Kati, line 120:
```suggestion
  assert_exit_code "1" "$(fake_function)"
```

Tito-Kati (approved):
I think it's fine, although I would consider incorporating the two small changes I've pointed out.

Issue #59, opened 2023-09-12
Display dots instead of tests names


Chemaclass:
## Current behaviour

Currently, the only output from bashunit is the test names in a pretty way.

## Acceptance Criteria

We want to allow rendering dots (a dot per test?). We can enable this feature when running bashunit with a `--dots` flag. Eg:

```bash
./bashunit tests/* --dots
```

PR #124, opened 2023-09-29
Output dots

Chemaclass:
## 📚 Description

Issue: https://github.com/TypedDevs/bashunit/issues/59: Display dots instead of tests names

Currently, the only output from bashunit is the test names in a `prettify way`. 
We want to allow rendering dots (a dot per test). 

## 🔖 Changes

- Render dots (per test) by default
- Render the scripts and test names using the new flag `--verbose`

Screenshot 2023-09-30 at 13 58 46

### DIFF
--- a/bashunit
+++ b/bashunit
@@ -35,7 +35,7 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
-    --verbose)
+    --verbose|-v)
       _VERBOSE=true
       shift
       shift
--- a/bashunit
+++ b/bashunit
@@ -24,6 +24,7 @@ source "$BASH_UNIT_ROOT_DIR/src/runner.sh"
 ###############
 
 _FILTER=""
+_DOTS=false
 _FILES=()
 
 while [[ $# -gt 0 ]]; do
@@ -32,6 +33,10 @@ while [[ $# -gt 0 ]]; do
     --filter)
       _FILTER="$2"
       shift
+      shift
+      ;;
+    --dots)
+      _DOTS=true
       shift
       ;;
     *)
--- a/bashunit
+++ b/bashunit
@@ -36,7 +36,8 @@ while [[ $# -gt 0 ]]; do
       shift
       ;;
     --dots)
-      _DOTS=true
+      _DOTS=${2-50}
+      shift
       shift
       ;;
     *)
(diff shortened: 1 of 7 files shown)


cmayo:
I like the output with the dots.

Why don't print the dots as default and show the current output with a verbose option? Maybe it has more sense to show the dots as a short info by default and if someone want more information then set the `--verbose` option. What do you think?

Chemaclass:
> Why don't print the dots as default and show the current output with a verbose option? 

Great idea, @cmayo ! Done ✔️  CC: @khru @Tito-Kati

bashunit
```diff
@@ -34,6 +35,11 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
+    --verbose)
```
cmayo, line 38: Just to allow also the short option name

```suggestion
    --verbose|-v)
```

Chemaclass: Oh, right! I thought about it, but I forgot it. Thanks!

Chemaclass:
We could think about doing in a follow up PR that if the test fail it will render an `f` instead of a `.` This will be a bit more complex than simply adding this:
```bash
# similar as we just did in this PR with `Console::printSuccessfulTest()`
function Console::printFailedTest() {
  ((_RENDERED_TEST_COUNT++))

  if [[ "$_VERBOSE" == false ]]; then
    if (( _RENDERED_TEST_COUNT % 50 != 0 )); then
      printf "f"
    else
      echo "f"
    fi
  else
    printf "..."
  fi
}
```

Because we should display at the end the actual test(s) that failed, and at this moment the only place where we display the error is inside `Console::printFailedTest()`

Tito-Kati (approved):

cmayo (approved):

Chemaclass:
> I like the output with the dots.
> 
> Why don't print the dots as default and show the current output with a verbose option? Maybe it has more sense to show the dots as a short info by default and if someone want more information then set the `--verbose` option. What do you think?

After some rethinking with @khru, we prefer keeping the "pretty" output by default and using the "dots" as optional. The reason is based on the value that you have on both options; when writing tests, the nicer version gives you already more value than the dots. I will create another PR to change this behaviour.

PR #126, opened 2023-09-30
Use verbose output by default

Chemaclass:
## 📚 Description

This reverts commit dc89a22014d52025729f7f88ac8145a263118f95.

## 🔖 Changes

Use "verbose output" by default, and allow using `--simple` to display a simplified output using a dot per passed test.

## 🖼️  Screenshots

Screenshot 2023-09-30 at 21 54 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [Unreleased](https://github.com/TypedDevs/bashunit/compare/0.6.0...main)
-- Added `--dots` argument for a simpler output
+- Added `--simple` argument for a simpler output
 - Manage error when test execution fails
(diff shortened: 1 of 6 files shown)


JesusValera (commented):
Not sure if like `--dots` as the argument name, maybe `--simple` would be easier/better?
Regarding the parameter, it should be fixed to 50 or 80 (check what's the default in PHPUnit for example), but giving the choice to the user to set this number doesn't provide value 🤔

Chemaclass:
Good idea, @JesusValera . I renamed `--dots` to `--simple`, and I removed the variable number, and fixed to `50`.

Tito-Kati (approved):

PR #128, opened 2023-10-01
Add some assert for file

khru:
## 📚 Description

- `assert_file_exists`
- `assert_file_not_exists`
- `assert_is_file_empty`
- `assert_is_file`

## ✅ To-do list

- [X] Make sure that all the pipeline passes
- [X] Make sure to update the `CHANGELOG.md` to reflect the new feature or fix

PR #129, opened 2023-10-01
Add assert for directory

cmayo:
## 📚 Description

Based on the new assert for files (https://github.com/TypedDevs/bashunit/pull/128), I have added new assert for directory:

  - `assert_directory_exists`
  - `assert_directory_not_exists`
  - `assert_is_directory`
  - `assert_is_directory_empty`
  - `assert_is_directory_not_empty`
  - `assert_is_directory_readable`
  - `assert_is_directory_not_readable`
  - `assert_is_directory_writable`
  - `assert_is_directory_not_writable`

## ✅ To-do list

### DIFF
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -249,3 +249,120 @@ function assert_is_file_empty() {
 
   State::addAssertionsPassed
 }
+
+function assert_directory_exists() {
+  local expected="$1"
+  local label="${2:-$(Helper::normalizeTestFunctionName "${FUNCNAME[1]}")}"
+
+  if [[ ! -d "$expected" ]]; then
+    State::addAssertionsFailed
+    Console::printFailedTest "${label}" "${expected}" "to exist but" "do not exist"
+    return
+  fi
+
+  State::addAssertionsPassed
+}
+
+function assert_directory_not_exists() {
+  local expected="$1"
+  local label="${2:-$(Helper::normalizeTestFunctionName "${FUNCNAME[1]}")}"
+
+  if [[ -d "$expected" ]]; then
+    State::addAssertionsFailed
+    Console::printFailedTest "${label}" "${expected}" "to not exist but" "the directory exists"
+    return
+  fi
+
+  State::addAssertionsPassed
+}
+
+function assert_is_directory() {
+  local expected="$1"
+  local label="${2:-$(Helper::normalizeTestFunctionName "${FUNCNAME[1]}")}"
+
+  if [[ ! -d "$expected" ]]; then
+    State::addAssertionsFailed
+    Console::printFailedTest "${label}" "${expected}" "to be a directory" "but is not a directory"
+    return
+  fi
+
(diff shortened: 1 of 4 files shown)


Tito-Kati:
@cmayo can you add the docs too please?

cmayo:
I'm having a problem while running the new tests on the windows container. It turns out the `chmod` command doesn't works on Windows and there are some tests failing on it. Should we skip those tests on Windows or does anyone know how to change the permissions from a bash script on Windows?

Tito-Kati:
@cmayo I don't have my Windows ready for dev, but I have one computer with Windows. If I get some free time tomorrow I'll take a look.

Tito-Kati:
@cmayo after a quick look to the PR, looks like this feature should be disabled on Windows (because Windows has different permission system). But let me check it anyways.

cmayo:
> @cmayo I don't have my Windows ready for dev, but I have one computer with Windows. If I get some free time tomorrow I'll take a look.

You can try it on linux with a this docker image:

`docker run -it -v .:/data mcr.microsoft.com/powershell`

Chemaclass:
💡 @cmayo , @Tito-Kati:  If you want to omit a test case for a certain platform you could use something like:
```bash
if [[ "$_OS" != "Windows" ]]; then
  # test case
fi
```
See `tests/unit/console_results_test.sh:128` or 144, 136, 155...

tests/unit/directory_test.sh
```diff
@@ -0,0 +1,180 @@
...
+    "$(Console::printFailedTest\
+      "Unsuccessful assert is directory readable when a file is given" \
+      "$a_file" "to be readable" "but is not readable")"\
+    "$(assert_is_directory_readable "$a_file")"
+}
+
+function test_unsuccessful_assert_is_directory_readable_without_execution_permission() {
+  if [[ "$_OS" == "Windows" ]]; then
+    return
+  fi
```
Chemaclass, line 112: I am ok with this for now. We can investigate why is it not working for Windows in a follow up iteration and check if it's even worth it or even possible to make this test compatible with Windows. CC: @Tito-Kati, @cmayo

src/assert.sh
```diff
@@ -249,3 +249,81 @@ function assert_is_file_empty() {
 
   State::addAssertionsPassed
 }
+
+function assert_directory_exists() {
+  local expected="$1"
+  local label="${3:-$(Helper::normalizeTestFunctionName "${FUNCNAME[1]}")}"
```
Chemaclass, line 255: Shouldn't this be using $2
```suggestion
  local label="${2:-$(Helper::normalizeTestFunctionName "${FUNCNAME[1]}")}"
```
Same applies to other assertions below where there is only one expected argument as $1, and the 2nd is the label.

cmayo:
I've added the inverse assertions for is empty, is readable and is writable.

Chemaclass (approved):

PR #138, opened 2023-10-03
Add --version

Chemaclass:
## 🔖 Changes

- Add --version option
- Add ASCII art logo

### simple

Screenshot 2023-10-03 at 11 55 14

### ascii art

Screenshot 2023-10-03 at 11 55 30


khru:
@Chemaclass This does not break the acceptance test?

Tito-Kati:
I'm not fully convinced by having bashunit written in ASCII (this is not friendly with screen readers). I'd prefer it to simply say "bashunit - 0.7.0" If you want, you can make the bashunit part bold to make it stand out more.

Anyone else with me on this?

Chemaclass:
I like your suggestion, @Tito-Kati , it's simpler in all possible ways. I will change it

khru:
> > I like your suggestion, @Tito-Kati , it's simpler in all possible ways. I will change it
> 
> imagen
> 
> The PHP one is very verbose, but it's still simple. We could add a link to the documentation if you find it too brief.

This could be one thing and the header other, I like both

khru (approved):
I've also change all the namespaces with Upper case to lowe case so they are all equal, added a test that should exist and separated the function to other file to ignor the .editorconfig rules

fabriziofs (approved):

khru (approved):
@Chemaclass I've added some changes to have default values and separate the logic of showing the header from showing the ascii

src/console_header.sh
```diff
@@ -0,0 +1,25 @@
...
+    printf "bashunit - %s\n" "$BASH_UNIT_VERSION"
+  fi
+}
+
+
+function console_header::print_version_with_env() {
+    local should_print_ascii="true"
+    if [[ "$SHOW_HEADER" != "$should_print_ascii" ]]; then
+      return
+    fi
```
Chemaclass, line 23: What about this, @khru ?
```suggestion
    if [[ "$SHOW_HEADER" != true ]]; then
      return
    fi
```

khru: The Shellcheck complains

Issue #139, opened 2023-10-03
Find all test on a folder


khru:
## Currently
The only way to call it is passing the name of the files to execute, but from a DevExp this sucks for example:

Trying to run all test from a folder:
![image](https://github.com/TypedDevs/bashunit/assets/6353105/4aa19644-77f5-4f9c-adc3-024033f51e20)

For example running all .sh of a folder, why should I filter all of them by a regex, should not the framework know what files to find and execute them?
![image](https://github.com/TypedDevs/bashunit/assets/6353105/324f7ca0-3d69-4fe4-b5c4-e55245e7ffef)

PR #140, opened 2023-10-03
Find all test on a directory

Chemaclass:
## 📚 Description

Issue: https://github.com/TypedDevs/bashunit/issues/139

## 🔖 Changes

- Allow running `./bashunit` over a full directory
- Update docs

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -100,7 +100,11 @@ Once **bashunit** is installed, you're ready to get started.
 
     You can copy and execute the following command from the root of your project if you installed **bashunit** as a Git submodule:
     ```bash
-    bashunit/bashunit tests/example_test.sh
+    # all tests inside the tests directory
+    ./bashunit tests
+
+    # or a concrete test by full path
+    ./bashunit tests/example_test.sh
     ```
 
 4.  If everything works correctly, you should see an output similar to the following:
--- a/src/helpers.sh
+++ b/src/helpers.sh
@@ -102,11 +102,11 @@ function helper::unsetIfExists() {
 }
 
 helpers::find_files_recursive() {
-    local path="$1"
+  local path="$1"
 
-    if [[ -d "$path" ]]; then
-        find "$path" -type f -name '*test.sh' | sort | uniq
-    else
-        echo "$path"
-    fi
+  if [[ -d "$path" ]]; then
+    find "$path" -type f -name '*test.sh' | sort | uniq
+  else
+    echo "$path"
+  fi
 }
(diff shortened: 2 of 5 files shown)


src/helpers.sh
```diff
@@ -100,3 +100,13 @@ function helper::unsetIfExists() {
...
+
+helpers::find_files_recursive() {
+    local path="$1"
+
+    if [[ -d "$path" ]]; then
+        find "$path" -type f -name '*test.sh' | sort | uniq
+    else
+        echo "$path"
+    fi
+}
```
Tito-Kati, line 112:
```suggestion
helpers::find_files_recursive() {
  local path="$1"

  if [[ -d "$path" ]]; then
    find "$path" -type f -name '*test.sh' | sort | uniq
  else
    echo "$path"
  fi
}
```

src/helpers.sh
```diff
@@ -100,3 +100,13 @@ function helper::unsetIfExists() {
     unset "$function_name"
   fi
 }
+
+helpers::find_files_recursive() {
+    local path="$1"
+
+    if [[ -d "$path" ]]; then
+        find "$path" -type f -name '*test.sh' | sort | uniq
```
Tito-Kati, line 108:
```suggestion
        find "$path" -type f -name '**/*test.sh' | sort | uniq
```

docs/getting-started.md
```diff
@@ -100,7 +100,11 @@ Once **bashunit** is installed, you're ready to get started.
 
     You can copy and execute the following command from the root of your project if you installed **bashunit** as a Git submodule:
     ```bash
-    bashunit/bashunit tests/example_test.sh
+    # all tests inside the tests directory
+    ./bashunit tests
```
Tito-Kati, line 104: I think the Test file names section should be updated too.

Tito-Kati (changes requested):

src/helpers.sh
Chemaclass: Not needed. The `find ... -name '*test.sh' ` is already looking at all files that has a name ending with `test.sh`, so we don't need to use the `**/*test.sh`. Or did you experience any issue testing it on Windows/Linux, @Tito-Kati ?

Tito-Kati: Only the ones directly under that directory, do we want recursive search?

Chemaclass: Isn’t `find` already searching recursively?
Screenshot 2023-10-03 at 15 18 27

JesusValera (approved):

src/helpers.sh
Tito-Kati: You are right xD

PR #155, opened 2023-10-07
Serve fonts from origin instead of cdn

### LINKED ISSUE #154: Serve the docs fonts ourselves
For the MVP of the documentation, fonts were served directly from Google's CDNs. The ideal scenario would be to serve the fonts ourselves.

Harsh-br0:
## 📚 Description

fixes #154

## 🔖 Changes

- added a `fonts.css` to link cached font files
- all cached fonts lies inside `docs/public/fonts` directory

## ✅ To-do list


Tito-Kati (changes requested):
Hello, thank you very much for helping us with the project, would you mind making some changes to simplify it?

Since the documentation is only in English, we do not need the versions with the unicode-ranges, the standard version of each font will be sufficient, you can download them from Google Fonts.

Make sure you download the correct weights for each font, we need 400 for Cousine, 600 for Rubik, and for Heebo we need both 400 and 600.

Additionally, it would be great if the filenames of the font files have a name that can be understood.

Thank you very much again for helping us.

Tito-Kati:
I also noticed that you've updated the yarn.lock, please @Harsh-br0 restore it to the previous version as we have it there precisely to ensure we all have the same versions of the packages and to ensure that the documentation works stably.

.gitignore
```diff
@@ -2,6 +2,10 @@
 .idea/
 .env
 
+# package manager
+package-lock.json
+yarn.lock
```
Chemaclass, line 7: We don't want to ignore these lock files

Harsh-br0: oh, let me fix it again

Harsh-br0:
Alright, my apologies for responding lately
so i have covered all the requested changes now

Harsh-br0:
Seems like my formatter messed up line endings

docs/.vitepress/theme/style.css
```diff
@@ -7,16 +7,25 @@
...
  */
 
-@import url('https://fonts.googleapis.com/css2?family=Heebo:wght@400;600&display=swap');
-@import url('https://fonts.googleapis.com/css2?family=Cousine');
-@import url('https://fonts.googleapis.com/css2?family=Rubik:wght@600');
+/**
+ * @import url('https://fonts.googleapis.com/css2?family=Heebo:wght@400;600&display=swap');
+ * @import url('https://fonts.googleapis.com/css2?family=Cousine');
+ * @import url('https://fonts.googleapis.com/css2?family=Rubik:wght@600');
+ */
```
Chemaclass, line 14: Are these lines needed? Why not removing them?

Harsh-br0: I kept them as a reference, but there seems no issue in removing them

Chemaclass: There is no need to keep commented code, especially if you're using **git** as it's a source version control and we can easily see the before/after and get to the removed code whenever needed (if needed) 😄

Chemaclass:
Can you please check the To-do list from the PR itself? 
At least the third box: "updating the CHANGELOG" is something you can also add to this PR :)

Screenshot 2023-10-08 at 18 09 34

Harsh-br0:
so, i have checked first 2 checkboxes which makes sense to this PR
But , should i commit the commit messages to `CHANGELOG.md` then the 3rd checkbox would be valid to this PR ?

Chemaclass:
@Harsh-br0, what I would suggest is to add  only one line like "Serve fonts from origin instead of cdn" to the Unreleased list from the CHANGELOG. That should be enough

Tito-Kati:
I believe that since it's not a significant improvement for our users, there's no need to add anything to the changelog, but it would be a shame if your name didn't appear in the next release.

Perhaps you could add something like "optimized docs fonts".

Harsh-br0:
Alright, Now it seems fine

Tito-Kati (approved):
Superb! Thanks you!

Harsh-br0:
Nice, Thanks for such amazing library.
It was good to work with you

PR #160, opened 2023-10-10
Improve the error msg when missing args

Chemaclass:
## 🔖 Changes

- Improve the error message when missing arguments 

### BEFORE
Screenshot 2023-10-10 at 22 02 25

### AFTER 

Screenshot 2023-10-10 at 22 01 16


Chemaclass:
I would rather create a --help option, and trigger that option in case no arguments are found.

PR #192, opened 2023-10-14
Define a custom .env when running bashunit

Chemaclass:
## 📚 Description

Issue: https://github.com/TypedDevs/bashunit/issues/187

## 🔖 Changes

- Add `--env` option
- Create and use `.env.test` in all acceptance tests

## ✅ To-do list


.env.test
```diff
@@ -0,0 +1,3 @@
+SHOW_HEADER=false
+HEADER_ASCII_ART=false
+SIMPLE_OUTPUT=false
```
khru, line 3: If I remember correctly, I think we commented that this .env should be in the acceptance test folder.

bashunit
```diff
@@ -40,6 +40,12 @@ while [[ $# -gt 0 ]]; do
       SIMPLE_OUTPUT=false
       shift
       ;;
+    --env)
+      # shellcheck disable=SC1090
+      source "$2"
+      shift
+      shift
+      ;;
```
khru, line 48: why 2 `shift`?

tests/acceptance/bashunit_test.sh
```diff
@@ -109,9 +109,9 @@ function test_bashunit_should_allow_test_drive_development() {
 
   set +e
 
-  assert_contains "$fixture_start" "$(./bashunit --verbose "$test_file")"
-  assert_contains "$fixture_end" "$(./bashunit --verbose "$test_file")"
-  assert_general_error "$(./bashunit --verbose "$test_file")"
+  assert_contains "$fixture_start" "$(./bashunit --env .env.test "$test_file")"
+  assert_contains "$fixture_end" "$(./bashunit --env .env.test "$test_file")"
+  assert_general_error "$(./bashunit --env .env.test "$test_file")"
```
khru: should the path to the `.env` file be on a variable an on a beforeEach?

bashunit
Chemaclass: 1st) for the -e|--env
2nd) for the value itself
> **Note**: You can see the filter using the same strategy.

khru (approved):

fabriziofs (approved):

PR #195, opened 2023-10-15
Fix and test installer

### LINKED ISSUE #194: The default behavior of the installer is not optimal (SRP)
The installer essentially has two behaviors: downloading a specific build based on its tag, and on the other hand, downloading and compiling the source code from the main or latest branch, the latter being its default behavior, as it is desirable for users to download the latest version unless they specify which version to install.

We know at any moment what the latest version is, we have several constants for the different environments that have it stored, so I believe that the latest (and therefore the default behavior) should act as an alias to that version that we know is the latest.

Especially emphasizing that the process that clones the repo is worse since, first it requires having dependencies installed like git, it surely downloads the entire project (including documentation, etc...) which is also slower.

Looking at it carefully and knowing that the installer is used to generate the builds for the releases, perhaps the real interesting thing would be to separate the script in such a way that the installer only stays with the second behavior and set up a new script for the builder that in fact wouldn't need to download the repository since it could do it with the local version (meaning we are currently not separating the responsibilities well in this script).

Tito-Kati:
## 📚 Description

closes #194

@ImanolRP was pairing with me during this pr

## 🔖 Changes

- Installer is not longer the builder
- Added acceptance tests to the installer 🎊 
- Little fix quickstart command

## ✅ To-do list

### DIFF
--- a/.github/RELEASE.md
+++ b/.github/RELEASE.md
@@ -3,13 +3,14 @@
 This is a guide to know the steps to create a new release.
 
 1. Update the version in [BASHUNIT_VERSION](../bashunit)
-2. Update the version in [CHANGELOG.md](../CHANGELOG.md)
-3. Update the version in [package.json](../package.json)
-4. Commit and push
-5. Create a [new release](https://github.com/TypedDevs/bashunit/releases/new) from GitHub
-6. Attach the latest executable to the release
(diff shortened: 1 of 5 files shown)


ImanolRP (approved):
💪

Chemaclass (approved):
Nice one! 👌👏

fabriziofs (approved):

PR #197, opened 2023-10-16
docs: add MacPorts install info

herbygillot:
## 📚 Description

Add instructions on how to install `bashunit` via MacPorts.
https://ports.macports.org/port/bashunit/

## 🔖 Changes

- Add entry to installation doc (`docs/installation.md`) detailing MacPorts installation of `bashunit`

## ✅ To-do list

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -37,6 +37,14 @@ You can install **bashunit** globally in your macOS (or Linux) using brew.
 brew install bashunit
 ```
 
+## MacPorts
+
+On macOS, you can also install **bashunit** via [MacPorts](https://www.macports.org):
+
+```bash
+sudo port install bashunit
+```
+
 ## Git submodule
 
 You can use Git submodules to include external Git repositories, like **bashunit**, within your Git project.


JesusValera:
Thank you so much for your PR @herbygillot 🙌🏼 

I just tried to install `macports` and `bashunits` and it works, but I have a question:
- Do we need to install such dependencies?
image

I would say most of them are not related to `bashunit`, additionally, it took a few minutes to be installed, although, if you try to install the program via `curl -s https://bashunit.typeddevs.com/install.sh | bash` it takes no time as the application is only 38k. 😃

herbygillot:
I think what's going on is that in `bashunit` depends on `bash` (we use MacPorts `bash` for consistency across macOS versions), and so all those dependencies are probably for `bash`. It's a chain of dependencies.

Chemaclass:
@herbygillot, what about new releases? Will this MacPort tool be updated automatically? Eg: https://github.com/macports/macports-ports/blob/master/devel/bashunit/Portfile#L6 or do we/you need to apply manual changes here? How do you usually work to keep up to date the tools under MacPorts? It's a new tool for me and I would like to understand it better before using it :)

Chemaclass:
> I think what's going on is that in `bashunit` depends on `bash` (we use MacPorts `bash` for consistency across macOS versions), and so all those dependencies are probably for `bash`. It's a chain of dependencies.

Actually, **bashunit** run perfectly (and it will keep this way for BC) on the default bash version that comes with MacOS, which is the 3.2 Screenshot 2023-10-16 at 22 30 49

I don't think we need to install the latest bash (and therefore that level of dependencies) when using `bashunit` in MacOS 🤔  For this reason, I think we could consider using the [installer tool](https://bashunit.typeddevs.com/installation#install-sh) that we created; it will ensure you have the latest version with only 38KB dependency that will run in your MacOS bash by default. What do you think, @herbygillot ?

herbygillot:
MacPorts follows a similar model to many other packaging projects (such as FreeBSD ports or Debian's) where every package has a maintainer, and it's the maintainer's responsibility to keep the package up to date.

A package in MacPorts is called a "port", and yes, I added and maintain the port for `bashunit`. You are correct that the procedure is updating the Portfile. I am subscribed to the repos for most all the software I maintain, and get updates that way, as well as through newreleases.io and Repology.

That aside, anyone can open a PR to update `bashunit`'s Portfile in the macports/macports-ports repo (or really any `Portfile` for any port, not just `bashunit`), and it will be reviewed and merged.

That said, we don't yet have automatic Portfile updates just yet, unfortunately.

herbygillot:
> Actually, **bashunit** run perfectly (and it will keep this way for BC) on the default bash version that comes with MacOS, which is the 3.2 Screenshot 2023-10-16 at 22 30 49
> 
> I don't think we need to install the latest bash (and therefore that level of dependencies) when using `bashunit` in MacOS 🤔 For this reason, I think we could consider using the [installer tool](https://bashunit.typeddevs.com/installation#install-sh) that we created; it will ensure you have the latest version with only 38KB dependency that will run in your MacOS bash by default. Or, another alternative, maybe it would be enough removing the line with `depends_lib-append port:bash` [L32 from here](https://github.com/macports/macports-ports/blob/master/devel/bashunit/Portfile#L32)? Would this be an option as [in brew we dont have bash as a dependency](https://github.com/Homebrew/homebrew-core/blob/3d22b36dd5c746ac6c510e72248a03a190d31357/Formula/b/bashunit.rb)... What do you think, @herbygillot ?

So the key point here is that MacPorts has an opposite philosophy to Homebrew: ports in MacPorts should depend as much as possible on other ports inside of MacPorts, and not so much on the external built-in macOS tooling.  This helps ensure that the MacPorts ecosystem is as self-contained as possible, and consistent across most versions of macOS, both old and new.  Installing `bash` along with `bashunit` isn't that big of a deal, and also many MacPorts users will have `bash` already installed, meaning that it only needs to fetch `bashunit`, which is a tiny download.

Chemaclass (approved):
Thanks for clarifying all open questions 💡🍻

herbygillot:
Thank you

Tito-Kati:
Thank you very much for your contribution @herbygillot , and thank you as well for taking the  time to resolve all our questions.

herbygillot:
> Thank you very much for your contribution @herbygillot , and thank you as well for taking the  time to resolve all our questions.

Thank you all for `bashunit` 🙏🏾

PR #199, opened 2023-10-18
Add assert_equals_ignore_colors

Chemaclass:
## 🔖 Changes

- Added `assert_equals_ignore_colors`

Screenshot 2023-10-18 at 10 38 52

## ✅ To-do list

### DIFF
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -14,7 +14,7 @@ function assert_equals() {
   state::add_assertions_passed
 }
 
-function assert_equals_without_colors() {
+function assert_equals_ignore_colors() {
   local expected="$1"
   local actual="$2"
   local label="${3:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -14,6 +14,20 @@ function assert_equals() {
   state::add_assertions_passed
 }
 
+function assert_equals_without_colors() {
+  local expected="$1"
+  local actual="$2"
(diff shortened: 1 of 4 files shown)


Tito-Kati:
@Chemaclass Would it be a good idea to change it to assert_equals_ignore_colors so it follows the same pattern as the ignore case?

Chemaclass:
Thanks, @Tito-Kati, I like `assert_equals_ignore_colors` much more, so I applied the renaming and pushed it.

Tito-Kati (approved):

src/assert.sh
```diff
@@ -14,6 +14,17 @@ function assert_equals() {
   state::add_assertions_passed
 }
 
+function assert_equals_ignore_colors() {
+  local expected="$1"
+  local actual="$2"
+  local label="${3:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
+
+  local actual_without_colors
+  actual_without_colors=$(echo -e "$actual" | sed "s/\x1B\[[0-9;]*[JKmsu]//g")
```
khru, line 23: would it make sense to give a better name to the regex that `sed` is ussing?

khru:
@Chemaclass would it make sense to use this on the acceptance test until we finish the snapshot testing?

Tito-Kati:
> @Chemaclass would it make sense to use this on the acceptance test until we finish the snapshot testing?

I think it is not worth it since the snapshot tests are almost ready

PR #202, opened 2023-10-21
Add docs for env config

Chemaclass:
## 🔖 Changes

- Add a new documentation page for env-vars

## 🖼️ Demo

https://github.com/TypedDevs/bashunit/assets/5256287/10049843-dffe-4322-9f5c-ef03d64fe4d2

## ✅ To-do list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add `assert_equals_ignore_colors`
 - Add `assert_match_snapshot`
 - Add `SHOW_EXECUTION_TIME` to environment config
+- Add docs for environment variables
 
 ## [0.9.0](https://github.com/TypedDevs/bashunit/compare/0.8.0...0.9.0) - 2023-10-15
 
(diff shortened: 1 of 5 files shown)


Tito-Kati (changes requested):
Could you put the defaults values too and change the texts to indicate what happens in case of being true? In the ones you refer to the command line, it would be nice to add what they do so you don't have to be jumping from one side to the other while reading the documentation.

And I think I would call the section just "Configuration".

The definition of the section could be a tad more commercial and explain a bit about its purpose, not so robotic, you could for example say that it serves to configure the behavior of bashunit in your project, that they need to create a .env file in the root directory but they can give it another name if they pass it as an argument to the command with --env (and a link to the corresponding section with more detail). I would also remove the example env or if you include it, provide one with values and not all arguments so that people can use it as an example.

In summary, if I read that documentation without any context currently, I wouldn't know how to configure bashunit.

Tito-Kati:
And I also think it's not necessary to put the bashunit command in every example, with the .env and the Output it's enough.

Tito-Kati:
Ah, and remember to update the command line section to say that they override the environment configuration and also add a link to the environment configuration if they wanted to make the change permanent.

Tito-Kati (approved):

Issue #152, opened 2023-10-07
Make the test dir entrypoint optional


Chemaclass:
## Current

Right now, you have to pass the directory|file of the test(s) that you want to run. Eg:
```bash
./basunit tests
```

## Expected

Allow defining the directory entry point as `.env` global variable. This will be the directory that will be taken if no directory or file was specified when running `.bashunit`. Eg:
```env
DEFAULT_PATH=tests
```

> DEFAULT_PATH is just a suggestion. Feel free to think about another better naming.

Miki471151:
Hello. I'd like to work on this issue. Please assign it to me.

Chemaclass:
@Miki471151, feel free to create a PR

Chemaclass:
@Miki471151, are you still interested on working on this feature?

Miki471151:
Yes. Im working on it.

Chemaclass:
@Miki471151, consider creating a draft PR to share the current progress so we can give you hints and ideas if needed. This helps to create transparency 🧠

PR #205, opened 2023-10-22
#152 - Make the test dir entrypoint optional

Miki471151:
## 📚 Description

The expected feature is to enable setting the directory entry point as a global variable in the .env file. This variable will serve as the default directory in case no directory or file is specified when running bashunit.

### **Executing ./bashunit without DEFAULT_PATH in .env:**
bashunit-nodefault

### **Executing ./bashunit with DEFAULT_PATH in .env:**
bashunit-default

### **Executing  ./bashunit with directory as argument.**
bashunit-arg

- Add .env variable `DEFAULT_PATH`.
- Add read_and_store_files_recursive function which recursively search for files within a specified directory and store the results in an array. 
- If no specific directory or file is defined when running the script, then it will use the stored value in the DEFAULT_PATH variable when it`s defined.

## ✅ To-do list


Chemaclass (approved):
Many thanks, @Miki471151!

Issue #114, opened 2023-09-25
Output of the duplicated test it's not understandable


khru:
#### Summary

Whenever you are developing a new test if you dupliate one the output it's impossible to understand

![image](https://github.com/TypedDevs/bashunit/assets/6353105/a19fc931-a6e3-4164-aac4-03537ead6bbf)

```
make test
Running tests/acceptance/bashunit_test.sh
✓ Passed: Bash unit when a test fail
✓ Passed: Bash unit when a test passes
Running tests/functional/logic_test.sh
✓ Passed: Other way of using the exit code
✓ Passed: Should validate a non ok exit code
✓ Passed: Should validate an ok exit code
✓ Passed: Text should be equal
✓ Passed: Text should contain
✓ Passed: Text should match a regular expression
✓ Passed: Text should not contain
✓ Passed: Text should not match a regular expression
Running tests/unit/assert_test.sh
✓ Passed: Successful assertArrayContains
✓ Passed: Successful assertArrayNotContains
✓ Passed: Successful assertCommandNotFound
✓ Passed: Successful assertContains
✓ Passed: Successful assertEmpty
✓ Passed: Successful assertEquals
✓ Passed: Successful assertExitCode
✓ Passed: Successful assertGeneralError
✓ Passed: Successful assertMatches
✓ Passed: Successful assertNotContains
✓ Passed: Successful assertNotEmpty
✓ Passed: Successful assertNotEquals
✓ Passed: Successful assertNotMatches
✓ Passed: Successful assertSuccessfulCode

✗ Failed: Successful fake
    Expected 'to be empty'
    but got '✗ Failed: Successful fake
    Expected '1'
    to be exactly '0''
✓ Passed: Successful return assertExitCode
✓ Passed: Unsuccessful assertArrayContains
✓ Passed: Unsuccessful assertArrayNotContains
✓ Passed: Unsuccessful assertCommandNotFound
✓ Passed: Unsuccessful assertContains
✓ Passed: Unsuccessful assertEmpty
✓ Passed: Unsuccessful assertEquals
✓ Passed: Unsuccessful assertExitCode
✓ Passed: Unsuccessful assertGeneralError
✓ Passed: Unsuccessful assertMatches
✓ Passed: Unsuccessful assertNotContains
✓ Passed: Unsuccessful assertNotEmpty
✓ Passed: Unsuccessful assertNotEquals
✓ Passed: Unsuccessful assertNotMatches
[...]

PR #206, opened 2023-10-22
#114 - Output of the duplicated test it's not understandable

adatamas:
## 📚 Description

Solution for the Issue#114 ( https://github.com/TypedDevs/bashunit/issues/114 ).  If it detects duplicate functions, it will print their names and the name of the file that contains them, with the following error message:
Screenshot_1
If several function names are duplciate in a file, they are also printed in a new line:
image

## 🔖 Changes

- Print the name of duplicate function names if one (in console_results.sh)
- Print the name of the file which contains duplicate function names if one (in console_results.sh)
- Add two new variables in the state.sh (`_FILE_WITH_DUPLICATED_FUNCTION_NAMES` and `_DUPLICATED_TEST_FUNCTIONS_FOUND`) with getters and setters
- Add a common setter that modifies all the 3 variables associated with the duplicated functions
- Add the necessary tests to verify the new features and changes mentioned above 

## ✅ To-do list


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Add `SHOW_EXECUTION_TIME` to environment config
 - Add docs for environment variables
 - Improve data provider output
+- Print the name of duplicate function names if one
+- Print the name of the file which contains duplicate function names if one
```
Chemaclass, line 13: What about this?
```suggestion
- Improve duplicated function names output
```

src/console_results.sh
```diff
@@ -7,6 +7,10 @@ function console_results::render_result() {
   if [[ "$(state::is_duplicated_test_functions_found)" == true ]]; then
     console_results::print_execution_time
     printf "%s%s%s\n" "${_COLOR_RETURN_ERROR}" "Duplicate test functions found" "${_COLOR_DEFAULT}"
+    # Print the filename which contains the duplicated functions
+    printf "File with duplicate functions: %s\n" "$(state::get_file_with_duplicated_function_names)"
+    # Print the list of duplicated functions
+    printf "Duplicate functions: %s\n" "$(state::get_duplicated_function_names)"
```
Chemaclass, line 13: Regardless of the logic (in this context), we could remove these comments because the code below each comment is self-explanatory.

src/state.sh
```diff
@@ -8,6 +8,8 @@ _ASSERTIONS_PASSED=0
 _ASSERTIONS_FAILED=0
 _ASSERTIONS_SKIPPED=0
 _ASSERTIONS_INCOMPLETE=0
+_DUPLICATED_FUNCTION_NAMES=""
+_FILE_WITH_DUPLICATED_FUNCTION_NAMES=""
```
Chemaclass, line 12: I am not sure if I like this to keep it globals, but I guess it would be OK as an MVP idea, and we could refactor it in a follow up if we see the opportunity.

src/helpers.sh
```diff
@@ -36,7 +36,7 @@ function helper::check_duplicate_functions() {
   local duplicates
   duplicates=$(echo "$sorted_names" | uniq -d)
   if [ -n "$duplicates" ]; then
-    state::set_duplicated_test_functions_found
+    state::set_duplicated_functions_merged "$script" "$duplicates"
```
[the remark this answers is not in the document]
adatamas: I just wanted to make it clear that this is a merged function, since the term duplicated function was already there by default, and I added two more global variables with similar names. Of course the modified version is also fine to me. Also, I noticed that the original code use the term 'duplicate functions' in some cases and 'duplicated functions'  in others, but I think their meaning is intended to be the same. It might be worthwhile to modify and standardize this as well.

Chemaclass (approved):
Thanks a lot, @adatamas! 
Any feedback from your side, @Tito-Kati or @khru?

Tito-Kati:
> Thanks a lot, @adatamas! Any feedback from your side, @Tito-Kati or @khru?

As an MVP it's fine, we'll iterate over this.

Tito-Kati:
Thanks you @adatamas !

khru:
> > Thanks a lot, @adatamas! Any feedback from your side, @Tito-Kati or @khru?
> 
> As an MVP it's fine, we'll iterate over this.

Same here

PR #212, opened 2023-10-30
Add unmock function

Chemaclass:
## 📚 Description

Currently, you can mock a particular function with custom logic, but there is no way to rollback the mock for other tests. It would be useful to be able to remove/rollback the mocked behaviour. 

## 🔖 Changes

- Add unmock function

All mocks will be removed/rollback to their original behaviour after each test. 
This means that the mock live scope is tied to the test scope

## ✅ To-do list

### DIFF
--- a/src/runner.sh
+++ b/src/runner.sh
@@ -136,6 +136,7 @@ function runner::run_test() {
     "$function_name" "$data" 2>&1 1>&3
 
     runner::run_tear_down
+    runner::clean_mocks
     state::export_assertions_count
   )
 
@@ -199,6 +200,12 @@ function runner::run_set_up_before_script() {
 
 function runner::run_tear_down() {
   helper::execute_function_if_exists 'tear_down'
+}
+
+function runner::clean_mocks() {
+  for i in "${!MOCKED_FUNCTIONS[@]}"; do
+    unmock "${MOCKED_FUNCTIONS[$i]}"
+  done
 }
 
 function runner::run_tear_down_after_script() {
--- a/src/runner.sh
+++ b/src/runner.sh
@@ -136,7 +136,7 @@ function runner::run_test() {
     "$function_name" "$data" 2>&1 1>&3
 
     runner::run_tear_down
-    runner::clean_mocks
+    runner::clear_mocks
     state::export_assertions_count
   )
 
@@ -202,7 +202,7 @@ function runner::run_tear_down() {
   helper::execute_function_if_exists 'tear_down'
 }
 
-function runner::clean_mocks() {
+function runner::clear_mocks() {
   for i in "${!MOCKED_FUNCTIONS[@]}"; do
     unmock "${MOCKED_FUNCTIONS[$i]}"
   done
--- a/tests/acceptance/mock_test.sh
+++ b/tests/acceptance/mock_test.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+#
+# Make sure that the runner::clean_mocks() is being called and removing the mocks and spies
+#
+function test_runner_clean_mocks_1() {
+  mock ls echo foo
+  assert_is_mock ls
+
+  spy ps
+  assert_is_mock ps
+}
+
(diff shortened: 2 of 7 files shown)


JesusValera (approved):

tests/unit/test_doubles_test.sh
```diff
@@ -24,6 +24,22 @@ EOF
...
+
+  mock ps<<EOF
+PID TTY          TIME CMD
+13525 pts/7    00:00:01 bash
+24162 pts/7    00:00:00 ps
+EOF
+
+  assert_equals "3" "$(helper::trim "$(code)")"
+  unmock ps
+  assert_not_equals "3" "$(helper::trim "$(code)")"
```
khru, line 40: How could return 3 on your OS?

khru: this has to still mock 
image

khru (changes requested):

tests/unit/test_doubles_test.sh
khru: an the test it's not telling you what behavior it's expected

Chemaclass: You're right. I will use a different approach.

src/runner.sh
```diff
@@ -127,6 +127,7 @@ function runner::run_test() {
     "$function_name" "$data" 2>&1 1>&3
 
     runner::run_tear_down
+    runner::clean_mocks
```
khru, line 130: Where are the test?

src/runner.sh
```diff
@@ -189,6 +190,12 @@ function runner::run_tear_down() {
   helper::execute_function_if_exists 'tear_down'
 }
 
+function runner::clean_mocks() {
```
khru, line 193: Where are the test

khru (changes requested):

src/runner.sh
Chemaclass: I added them to `tests/acceptance/mock_test.sh`

khru: we are also missing unit test of the function, no?

khru, line 130: In other frameworks, this function is call `clear_mocks`
image

tests/acceptance/mock_test.sh
```diff
@@ -0,0 +1,17 @@
...
+  assert_is_mock ls
+
+  spy ps
+  assert_is_mock ps
+}
+
+function test_runner_clean_mocks_2() {
+  assert_is_not_mock ls
+  assert_is_not_mock ps
+}
```
khru, line 17: These test are not acceptance test, and the behavior you are trying to test it's not clear

tests/unit/test_doubles_test.sh
```diff
@@ -24,6 +24,25 @@ EOF
...
+13525 pts/7    00:00:01 bash
+24162 pts/7    00:00:00 ps
+EOF
+
+  assert_is_mock ps
+  unmock ps
+  assert_is_not_mock ps
+}
+
+function test_successful_unmock_spy() {
```
khru, line 39: where are the unsuccessful test for this?

khru (changes requested):

Tito-Kati:
@Chemaclass this is more complex than we anticipated, do you mind if we leave it in draft until we can get together? Besides, I think it's more priority to refactor the acceptance tests and finish the snapshots.

tests/unit/test_doubles_test.sh
Chemaclass: Added

src/runner.sh
Chemaclass: Could you give me a hint how would you write unit tests for `runner::clear_mocks()`?

Chemaclass: Done

CosmeValera (approved):

Chemaclass:
Close because inactivity. I will open it again to seek for new feedback

PR #218, opened 2023-11-08
Snapshots iteration 2

Tito-Kati:
## 📚 Description

This is the second iteration on the snapshot functionality. With this iteration, and in my opinion, I believe this functionality could be released.

Now when the snapshots are created (executing snapshot assertion the first time), the test is not considered as having passed. Instead, they are marked with the 'snapshot' status and the user is appropriately informed that these snapshots have been created.

image

## ✅ To-do list


Chemaclass (approved):
Kudos to you, @Tito-Kati ! I love it

khru (approved):
image

PR #222, opened 2023-11-12
feat: added upgrade option

gpuligundla:
## 📚 Description

## 🔖 Changes

- List individual changes in more detail

## ✅ To-do list

### DIFF
--- a/src/helpers.sh
+++ b/src/helpers.sh
@@ -153,3 +153,36 @@ function helper::trim() {
 
     echo "$trimmed_string"
 }
+
+function helpers::get_latest_tag() {
+    local remote_url="https://github.com/TypedDevs/bashunit"
+    local latest_tag
+
+    latest_tag=$(git ls-remote --tags "$remote_url" | awk '{print $2}' |
+    sed 's|^refs/tags/||' | sort -V | tail -n 1
+    )
+
+    echo "$latest_tag"
+}
+
+function helpers::upgrade() {
+    local script_path
+    local latest_tag
+
+    script_path="$(realpath "$BASHUNIT_ROOT_DIR")"
+    latest_tag="$(helpers::get_latest_tag)"
+
+    if [[ $BASHUNIT_VERSION == "$latest_tag" ]]; then
+      echo "> You are already on latest release."
+      return
+    fi
+
+    echo "> Upgrading bashunit to latest release."
+    git clone --depth 1 --no-tags -b latest https://github.com/TypedDevs/bashunit /tmp/bashunit 2>/dev/null
+    cd /tmp/bashunit || exit
+    ./build.sh >/dev/null
+    mv "$script_path/bashunit" "$script_path/bashunit.old"
+    cp ./bin/bashunit "$script_path/bashunit"
+    rm -f "$script_path/bashunit.old"
+    echo "> bashunit upgraded successfully to latest version $BASHUNIT_VERSION."
+}
--- a/bashunit
+++ b/bashunit
@@ -54,6 +54,10 @@ while [[ $# -gt 0 ]]; do
       console_header::print_version
       trap '' EXIT && exit 0
       ;;
+    --upgrade)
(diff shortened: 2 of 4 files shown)


src/helpers.sh
```diff
@@ -153,3 +153,36 @@ function helper::trim() {
...
+
+function helpers::get_latest_tag() {
+    local remote_url="https://github.com/TypedDevs/bashunit"
+    local latest_tag
+
+    latest_tag=$(git ls-remote --tags "$remote_url" | awk '{print $2}' |
+    sed 's|^refs/tags/||' | sort -V | tail -n 1
+    )
+
+    echo "$latest_tag"
```
Chemaclass, line 165: Would this work out as well, without the need for a local variable?
```suggestion
  git ls-remote --tags "$remote_url" |
    awk '{print $2}' |
    sed 's|^refs/tags/||' |
    sort -V |
    tail -n 1
```

Tito-Kati:
Thank you very much @gpuligundla, what I see looks very promising.

Just need to add some tests, include you in the changelog.md, and add the new option to the documentation (it would go here: docs/command-line.md).

khru (changes requested):
I Agree with @Tito-Kati I'm lacking some acceptance test and some unit test for this

src/helpers.sh
gpuligundla: changed it

tests/acceptance/upgrade_bashunit_test.sh
```diff
@@ -0,0 +1,29 @@
...
+    local output
+
+    output="$(./bin/bashunit --upgrade)"
+
+    assert_equals\
+    "$(printf "> You are already on latest release.")" "$output"
+}
+
+function test_fake_bashunit_upgrade() {
+    sed -i 's/declare -r BASHUNIT_VERSION="[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}"/declare -r \
```
Chemaclass, line 22: This will fix the broken test on macOS
```suggestion
    sed -i -e 's/declare -r BASHUNIT_VERSION="[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}"/declare -r \
```

tests/unit/helpers_test.sh
```diff
@@ -190,3 +190,7 @@ function test_find_files_recursive_given_wildcard() {
 
   assert_equals "tests/unit/fixtures/tests/example2_test.sh" "$result"
 }
+
+function test_get_latest_tag() {
+  assert_not_empty "$(helpers::get_latest_tag)"
+}
```
Chemaclass, line 196: Calling the actual method will imply I/O which by definition is not a unit test, because you are depending on external things. Therefore, we can mock the network response, and we will test the actual behaviour of the function. Consider the following example:
```suggestion
function test_get_latest_tag() {
  mock git<<EOF
fc9aac40eb8e5ad4483f08d79eb678a3650dcf78        refs/tags/0.1.0
a17e6816669ec8d0f18ed8c6d5564df9fc699bf9        refs/tags/0.10.0
3977be123b0b73cfdf4b4eff46b909f37aa83b3c        refs/tags/0.10.1
b546c693198870dd75d1a102b94f4ddad6f4f3ea        refs/tags/0.2.0
732ea5e8b16c3c05f0a6977b794ed7098e1839e2        refs/tags/0.3.0
EOF

  local actual
  actual="$(helpers::get_latest_tag)"

  assert_equals "0.10.1" "$actual"
  unset -f git
}
```

gpuligundla: yeah, changed it. I never worked with the concept of mock. Thanks for introducing it to me.

Chemaclass (approved):
Awesome! Thanks a lot 💯

Tito-Kati (approved):
Thank you for collaborating with us and adding this great feature to our library.

PR #228, opened 2023-11-18
Optimize bashunit

Tito-Kati:
## 📚 Description

By removing the call to the deprecated setUp/tearDown functions and modifying the implementation of the execute_function_if_exists and unset_if_exists helpers, I've managed to reduce the execution time of the library's own tests by around 15% in my local environment.

To calculate these times, I have been running the following script while making changes and taking an average of the times:

```bash
for i in {1..10}; do make test | grep "Time taken: " >> result_2_f; done
```

After all the changes, the results I have obtained from main vs this branch are as follows:
imagen

Removing the deprecations meant a reduction of about 5%, and modifying the implementation of the helpers accounted for the remaining 10% reduction.

## 🔖 Changes

- Remove support to deprecated setUp/tearDown functions
- Optimize test execution speed

## ✅ To-do list

### DIFF
--- a/src/helpers.sh
+++ b/src/helpers.sh
@@ -85,22 +85,14 @@ function helper::get_functions_to_run() {
 # @param $1 string Eg: "do_something"
 #
 function helper::execute_function_if_exists() {
-  local function_name=$1
-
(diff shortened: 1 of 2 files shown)


Chemaclass (approved):
I love it! Kudos!

PR #235, opened 2023-12-10
unmock all test doubles after each test

Chemaclass:
## 📚 Description

Currently, you can mock a particular function with custom logic, but there is no way to rollback the mock for other tests. 
The goal is to be able to rollback the mocked behaviour to the original one - and this should happen automatically after each test case.

## 🔖 Changes

- Add `unmock`
- Call `clear_mocks` after `run_tear_down` for each test

## ✅ To-do list

### DIFF
--- a/docs/test-doubles.md
+++ b/docs/test-doubles.md
@@ -27,7 +27,7 @@ Allows you to override the output of a callable.
 ```bash [Example]
 function test_example() {
   function code() {
-      ps a | grep bash
+    ps a | grep bash
   }
 
   mock ps<<EOF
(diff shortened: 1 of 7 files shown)


CosmeValera (approved):

fabriziofs (approved):

docs/test-doubles.md
```diff
@@ -25,7 +27,31 @@ Allows you to override the output of a callable.
...
+13525 pts/7    00:00:01 bash
+24162 pts/7    00:00:00 ps
+EOF
+
+  assert_equals "13525 pts/7    00:00:01 bash" "$(code)"
+}
+```
+:::
+
+## unmock
```
antonio-gg-dev, line 44: In my honest opinion, I wouldn't document the unmock as it could cause confusion among our users. The option is there, but the unmock should be automatic. Documenting it might lead our users to think they need to perform this process manually.

antonio-gg-dev (approved):

docs/test-doubles.md
Chemaclass: Ok, agreed. I will simply remove it from the docs - and it will be like an easter egg feature. You can use it in your tests, but you dont have to, as it will happen automatically after each test anyway.

PR #239, opened 2023-12-21
Support Bash "strict mode" (and fix one unit test)

djpohly:
## 📚 Description

Some script authors like to use ["unofficial Bash strict mode"](http://redsymbol.net/articles/unofficial-bash-strict-mode/), enough so that supporting it shows up in [comparisons of shell testing frameworks](https://github.com/dodie/testing-in-bash#detailed-comparision).

Case in point: enabling the -u option uncovered a bug in `test_unsuccessful_assert_is_directory_not_writable` that is fixed in this PR.

## 🔖 Changes

- bashunit and its tests now have `set -euo pipefail` enabled.
- The test mentioned above has been fixed (`$a_file` was supposed to be `$a_directory`).
- Arithmetic expansions `((...))`, which exit 1 if their result happens to be zero, are fixed with `|| true`.
- The call to `check_duplicate_functions` does not appear to be intended to cause the script to fail, so it is also silenced with `|| true`.
- Optional parameters are given default-empty behavior using e.g. `${1-}`.
- Optional variables are assigned defaults by evaluating `"${varname=default}"`.
- Existence of optional functions is tested explicitly before attempting to execute them.  This has the side benefit of ensuring that an external program with the right name (e.g. `set_up`) will not be run instead.

## ✅ To-do list

- [ ] ~~I updated the documentation to reflect the changes~~ (Not sure there's anything to document, since this fix doesn't require users to do anything different.)


Chemaclass (approved):
Wow, this is pretty cool. Thanks!

Chemaclass:
It seems that it is not working as expected for macos... I will check it out when I find some time these days

Chemaclass:
I will merge this branch into an internal one with the same name `strictmode`, so I can check it out and play around to reproduce the bug easier, while keeping the original commits from @djpohly 👍🏼

PR #241, opened 2023-12-24
Foundation for strictmode

Chemaclass:
## 📚 Description

Original PR: https://github.com/TypedDevs/bashunit/pull/239 by @djpohly  

Some script authors like to use ["unofficial Bash strict mode"](http://redsymbol.net/articles/unofficial-bash-strict-mode/), enough so that supporting it shows up in [comparisons of shell testing frameworks](https://github.com/dodie/testing-in-bash#detailed-comparision).

Case in point: enabling the -u option uncovered a bug in `test_unsuccessful_assert_is_directory_not_writable` that is fixed in this PR.

## 🔖 Changes

- The test mentioned above has been fixed (`$a_file` was supposed to be `$a_directory`).
- Arithmetic expansions `((...))`, which exit 1 if their result happens to be zero, are fixed with `|| true`.
- The call to `check_duplicate_functions` does not appear to be intended to cause the script to fail, so it is also silenced with `|| true`.
- Optional parameters are given default-empty behavior using e.g. `${1-}`.
- Optional variables are assigned defaults by evaluating `"${varname=default}"`.

## 🏗️  Follow up

- Add `set -euo pipefail` to all tests. Currently not possible; needs some more investigation [wip]

## ✅ To-do list


Chemaclass:
Adding `set -euo pipefail` is not possible yet due to some unknown misbehaviours(🐛 ?)
I will rather work in a follow up PR the fix for it, enabling us to merge the changes of this PR. 

For example, `./bashunit tests/acceptance/mock_test.sh` without debugging mode (`set -x`):
Screenshot 2024-07-12 at 14 48 38

And with the debug mode enabled we can already see something, but still is not clear how to fix it:
Screenshot 2024-07-12 at 14 51 06

fabriziofs (approved):

staabm:
no ideas on my end to this - sorry.

PR #242, opened 2024-01-12
fix: use the bash executable specified by PATH

apotterri:
## 📚 Description

Update the release script so the final bashunit script will use PATH to find bash. The version of bash in /bin on macOS is very old, and forcing tests to use it is quite limiting.

## 🔖 Changes

- Change `build.sh` so it prepends `#!/usr/bin/env bash` to the final `bashunit` script.

## ✅ To-do list

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,9 @@
 
 mkdir -p bin
 
-cat src/*.sh > bin/temp.sh
+echo '#!/usr/bin/env bash' > bin/temp.sh
+
+cat src/*.sh >> bin/temp.sh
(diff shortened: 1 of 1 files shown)


apotterri:
This seems like a workable approach, but I'm not sure it's best (especially because there don't appear to be any tests that use the final packaged `bashunit` script).

Happy to get any feedback on this change in general as well.

Chemaclass:
Everything seems to be working fine. Thanks for the PR, @apotterri !

Screenshot 2024-01-13 at 17 07 24

Chemaclass (approved):

PR #243, opened 2024-01-15
fix: run test functions in defined order

### LINKED ISSUE #217: Test execution order
#### Summary

Test functions within a test file are executed (or at least their results are printed) alphabetically, regardless of the order in which you define them.

For me, it would make more sense for these to be executed in the order they are defined. For example, if you are doing TDD with the stop on failure option active, and you are adding tests one after another, it may be that if your last test fails (which is very common in TDD), it could be hiding the fact that you broke one of the tests you had already developed.

#### Current behavior

Test functions are executed (or at least their results are printed) alphabetically.

#### How to reproduce

This can be seen by creating a test file in which the test functions are not arranged alphabetically.

#### Expected behavior

Test functions should be executed and printed in the order they are defined within the test file.

apotterri:
Fixes #217.

## 📚 Description

Run the test functions in the order they're defined in the test file.

## ✅ To-do list


Chemaclass (approved):
Interesting idea, I didn't think about it. What do you think, @antonio-gg-dev and @khru ?

khru:
O think that test must be independent

https://kentbeck.github.io/TestDesiderata/

antonio-gg-dev (approved):

PR #245, opened 2024-02-26
Increase contrast of test results 

h0adp0re:
## 📚 Description

Improve test report contrast.

### Before
Screenshot 2024-02-26 at 19 17 04

### After
Screenshot 2024-02-26 at 19 37 38

## 🔖 Changes

- Improve the final test result state contrast.
- Add whitespace to the final test result state.
- Fix "filer" -> "filter" typo in the help command.
- Simplify `.editorconfig` to specify as little as possible, making the file more readable.
  The root cause of this change was neovim not respecting the defined indents for `.sh` files.

## ✅ To-do list

- [ ] I updated the documentation to reflect the changes (N/A?)


src/assert.sh
```diff
@@ -20,7 +20,7 @@ function assert_equals_ignore_colors() {
   local label="${3:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
 
   local actual_without_colors
-  actual_without_colors=$(echo -e "$actual" | sed "s/\x1B\[[0-9;]*[JKmsu]//g")
```
h0adp0re, line 23: This change was needed because of `tput sgr0`:
https://stackoverflow.com/a/35582778

src/assert.sh
```diff
@@ -20,7 +20,7 @@ function assert_equals_ignore_colors() {
...
+  actual_without_colors=$(echo -e "$actual" | sed "s/\x1B\[[0-9;]*[JKmsu]//g" | sed "s/\x0f//g")
```
h0adp0re, line 23: This change was needed because of `tput sgr0`:
https://stackoverflow.com/a/35582778

h0adp0re:
And another suggestion that I did not yet implement.
What if the final result state had padding whitespace?
Screenshot 2024-02-26 at 21 54 07

tests/acceptance/bashunit_execution_error_test.sh
```diff
@@ -8,16 +8,14 @@ function test_bashunit_when_a_execution_error() {
   local test_file=./tests/acceptance/fixtures/test_bashunit_when_a_execution_error.sh
   local fixture_start
   fixture_start=$(printf "Running ./tests/acceptance/fixtures/test_bashunit_when_a_execution_error.sh
-\e[31m✗ Failed\e[0m: Error
-    \e[2mExpected\e[0m \e[1m\'127\'\e[0m
-    \e[2mto be exactly\e[0m \e[1m\'1\'\e[0m
-\e[31m✗ Failed\e[0m: Error
-    \e[2m./tests/acceptance/fixtures/test_bashunit_when_a_execution_error.sh:")
+$(tput setaf 1)✗ Failed$(tput sgr0): Error
```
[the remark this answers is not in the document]
h0adp0re: I wrote something like this, let me know if that's acceptable:
```bash
  fixture_start=$(
    printf "Running ./tests/acceptance/fixtures/test_bashunit_when_a_execution_error.sh"
    printf "\n%s%s" \
      "${color_red}" "✗ Failed"
    printf "%s%s" \
      "${color_normal}" ": Error"
    printf "\n    %s%s%s" \
      "${color_dim}" "Expected" "${color_normal}"
    printf " %s%s%s" \
      "${color_bold}" "'127'" "${color_normal}"
    printf "\n    %s%s%s" \
      "${color_dim}" "to be exactly" "${color_normal}"
    printf " %s%s%s" \
      "${color_bold}" "'1'" "${color_normal}"
  )
  fixture_end=$(
    printf "\n%s%s%s" \
      "${color_dim}" "Tests:     " "${color_normal}"
    printf " %s%s%s%s" \
      "${color_red}" "1 failed" "${color_normal}" ", 1 total"
    printf "\n%s%s%s" \
      "${color_dim}" "Assertions:" "${color_normal}"
    printf " %s%s%s%s" \
      "${color_red}" "1 failed" "${color_normal}" ", 1 total"
  )
```

Chemaclass:
Thanks for your effort, @h0adp0re! Can you please take care of the broken CI steps?

h0adp0re:
Thanks for the library! Sure, it's the `printf` variable warning, I'll just commit it right away then.

h0adp0re:
> And perhaps even a preceding newline? Screenshot 2024-02-26 at 22 01 21

I implemented this version with the latest commit.

antonio-gg-dev:
More broken tests on the CI @h0adp0re 🙏🏻

h0adp0re:
Hmm, I  will have to think about this. I thought I mitigated the missing `$TERM` problem for `tput` [here](https://github.com/TypedDevs/bashunit/compare/1ad662c92d81dcc024e78fd70a6c846d7e5adde5..263381976c32484dc75f4451c3425fc257063740).

h0adp0re:
@Chemaclass @antonio-gg-dev @khru 
I've reached a conclusion that with `tput` the snapshots _will_ be platform-specific. I don't see a way around it, unless a decision is made that the snapshots don't contain any escape codes.

If no such decision can be made at this time, I will have to revert the migration to `tput` and only stick with the color and spacing changes.

antonio-gg-dev:
My vote is for keeping the behavior of bashunit as consistent across environments as possible, so I would only remove the use of tput while maintaining the changes in colors and spacing as you indicate.

On the other hand, omitting the escape codes from the snapshots seems like an interesting feature, but verbosely, as an alternative type of snapshot assertion, I would always maintain that the default behavior includes them.

h0adp0re:
Yes, everything you said makes sense. I am up for adding `assert_match_snapshot_ignore_colors`, if this feature would interest you. Thus, I see three options at the moment:

1. I will revert this project back to hard-coded escape sequences, and:
   1. the PR will be done,
   2. add `assert_match_snapshot_ignore_colors`.
2. I will add `assert_match_snapshot_ignore_colors` and implement it in the project's snapshot tests, making snapshots independent of colors.

My order of preference is `2`, `1.2`, `1.1`.

antonio-gg-dev:
Hi @h0adp0re !

For now, work on 1.i so we can close this PR and keep it more concise with a single improvement.

You can work on 1.ii next if you want, but as I said, we want bashunit to operate well-covered and consistent across different environments, so even with this new assertion, it's best if our acceptance tests don't use it.

If tput is so inconsistent across environments, to me this means discarding its use.

h0adp0re:
Hey, thanks for confirming. I took a small break from this problem to process it properly and I agree, I will do that. Atomic changes are traceable and maintainable. And, in the end, all I wanted to achieve was the contrast change.

I might take on `assert_match_snapshot_ignore_colors` in the near future when I get to testing some colorful scripts.

And at the risk of sounding repetitive — I still do recommend using `tput` in any script that is expected to be run in any number of unknown shells. Its variability is a deliberate feature that ensures the correct escape sequences for each environment. But for general use, I have to admit, `ESC[` … `m` is _probably_ supported widely enough.

antonio-gg-dev:
Thank you very much for understanding @h0adp0re , of course, take all the time you need and don't feel pressured to do it. If at any point you decide not to continue, just let us know and we will take care of it from there.

We greatly appreciate the time you're dedicating to bashunit and, of course, we're going to consider the use of tput (or an alternative) now that we're aware of the problem we have and the advantages of using this tool.

But as I mentioned before, we're not comfortable making large changes that have a significant impact on bashunit overall (and this is evident by the number of tests that fail or would need to be changed), so for now, this is the reason why we prefer to set it aside for a more in-depth investigation.

I have created these 2 issues (#246 and #247) to not lose the progress that has emerged from this PR and to know where to continue.

h0adp0re:
Thanks for the issues, I can do the changes tomorrow.

tests/acceptance/bashunit_execution_error_test.sh
h0adp0re, line 1: Is this change necessary? It does make the test more readable, but the use of variables _may_ defeat the purpose of the test.

antonio-gg-dev: I agree that using app global variables removes part of the test's purpose, but it's as simple as redeclaring them locally (with different names) within the test and not using the application's global ones.

antonio-gg-dev (approved):

tests/acceptance/bashunit_execution_error_test.sh
h0adp0re: I think it's all done now.

antonio-gg-dev (approved):

antonio-gg-dev:
Thank you very much @h0adp0re for contributing to bashunit!

Issue #225, opened 2023-11-15
Remove downtime on install latest


Chemaclass:
## Current

When creating a new release, we have to change the release version in many files (see `.github/RELEASE.md`). However, currently the `install.sh` is one of them and that's creating a downtime when we are creating a new release - during the time when this change it's pushed and the actual new release is created, the installer tool wont work.

## Expected

Instead of using a hardcoded version on `install.sh`, we should get the latest stable version.
In this [PR](https://github.com/TypedDevs/bashunit/pull/222/files), we introduced `get_latest_tag()` and we could use it on `install.sh` script instead of a hardcoded version.

```bash
function helpers::get_latest_tag() {
  git ls-remote --tags "$BASHUNIT_GIT_REPO" |
    awk '{print $2}' |
    sed 's|^refs/tags/||' |
    sort -V |
    tail -n 1
}
```

gpuligundla:
Can i work on it ?

Tito-Kati:
Great to see you back here again, @gpuligundla !

Of course, would you mind waiting a bit before starting it? We want to make some small changes to your last contribution that might affect the way to solve this issue.

When we make the changes, we'll let you know so that you can take a look at them too and then you can start this one. Meanwhile, I'll go ahead and assign it to you. 😊

Tito-Kati:
@gpuligundla we already merged #227 so you can start this issue when you want 😃

Chemaclass:
Actually, I think we should copy/paste the function `helpers::get_latest_tag()` function entirely in the install.sh script, because it must be a fully independent script, as it it what the users will download (as single script) to install bashunit. So we cannot source/load other scripts within the installer.

maheshmeleti:
I would like to work on this, please assign this issue to me.

Chemaclass:
@maheshmeleti let us know if you need help with this

maheshmeleti:
@Chemaclass im still working on it

PR #248, opened 2024-03-08
added get_latest_tag in install.sh

maheshmeleti:
## 📚 Description

Remove downtime on install latest #225

## 🔖 Changes

1. Updated install.sh to fetch the latest version from the git hub instead of hard coding.

2. Updated RELEASE.md file.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,12 +2,22 @@
 
 # shellcheck disable=SC2164
 # shellcheck disable=SC2103
+declare -r BASHUNIT_GIT_REPO="https://github.com/TypedDevs/bashunit"
 
-declare -r LATEST_BASHUNIT_VERSION="0.11.0"
+function get_latest_tag() {
+  git ls-remote --tags "$BASHUNIT_GIT_REPO" |
+    awk '{print $2}' |
+    sed 's|^refs/tags/||' |
+    sort -Vr |
+    head -n 1
+}
+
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -2,13 +2,23 @@
 
 # shellcheck disable=SC2164
 # shellcheck disable=SC2103
+declare -r BASHUNIT_GIT_REPO="https://github.com/TypedDevs/bashunit"
+
+function get_latest_tag() {
```
Chemaclass, line 7: Can you write a test to ensure the behaviour of this new function?

Chemaclass (commented):
Thanks for your PR! Although, I am missing some tests for this new logic. Some ideas: an acceptance test to make sure the whole thing works as expected, and/or a unit test for the new little function behaviour.

install.sh
maheshmeleti: I am using the existing function for the latest update; 
A test case already exists to test the function test_get_latest_tag() in \bashunit\tests\unit\helpers_test.sh.

khru (approved):
Looks good

Chemaclass (approved):

install.sh
Chemaclass: You are right, that test is covering that area as well, thanks!

antonio-gg-dev (approved):

antonio-gg-dev:
Closes #225

PR #249, opened 2024-04-24
chore: add fail() function

ceejatec:
## 📚 Description

Adds new `fail()` function for general purpose test failures not covered by the various `assert_*` functions.

## 🔖 Changes

- Added `fail()` function in assert.sh
- Added `console_results::print_failure_message()` function in console_results.sh
- Added test case

## ✅ To-do list

### DIFF
--- a/docs/assertions.md
+++ b/docs/assertions.md
@@ -856,3 +856,20 @@ function test_failure() {
 }
 ```
 :::
+
+## fail
+> `fail "failure message"`
+
+Unambiguously reports an error message. Useful for reporting specific message
+when testing situations not covered by any `assert_*` functions.
+
+::: code-group
+```bash [Example]
+function test_success() {
+  true || fail "This will never fail"
+}
+function test_failure() {
+  [ $(date +%-H) -lt 6 ] || fail "It's late, go to bed"
+}
+```
+:::
--- a/docs/assertions.md
+++ b/docs/assertions.md
@@ -866,10 +866,14 @@ when testing situations not covered by any `assert_*` functions.
 ::: code-group
 ```bash [Example]
 function test_success() {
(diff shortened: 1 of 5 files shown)


Chemaclass (approved):
How interesting. Is this something you have seen in other libraries, @ceejatec?

docs/assertions.md
```diff
@@ -856,3 +856,20 @@ function test_failure() {
...
+when testing situations not covered by any `assert_*` functions.
+
+::: code-group
+```bash [Example]
+function test_success() {
+  true || fail "This will never fail"
+}
+function test_failure() {
+  [ $(date +%-H) -lt 6 ] || fail "It's late, go to bed"
+}
```
antonio-gg-dev, line 873: The example you set for test_failure doesn't always fail because it depends on the time of execution. The idea is to have one of the examples always succeed and the other always fail. Could you change it?

I think it would also help to use the if syntax instead of the || operator in both examples, as it's more familiar to a greater number of programmers.

antonio-gg-dev (changes requested):
Thank you very much for your contribution and for treating bashunit with the same care as we do, I love your idea.

ceejatec:
@Chemaclass Indeed, it's a pretty common function, going [right back to JUnit](https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#fail(java.lang.String)). IME, frequently the various assert() methods are internally implemented in terms of the basic `fail(message)` function too.

docs/assertions.md
ceejatec: I've modified them as requested. I had trouble coming up with an example that could seem useful in only a line or two, but I think this gets the job done.

antonio-gg-dev (approved):
Thank you!

PR #253, opened 2024-05-24
fix: `bashunit-e` file isn't created sometimes

fabriziofs:
## 📚 Description

Fix failing test due to the way `sed` command behave in macOS, sometimes the file `./bin/bashunit-e` isn't created  and the test fails:
image

To avoid this I've added the `-f` flag to the `rm` command to in case the `./bin/bashunit-e` don't exist the command don't fail and breaks the test, doc related:
image

## 🔖 Changes

- Add `-f` flag to avoid problems with `rm` command in macOS

## ✅ To-do list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased](https://github.com/TypedDevs/bashunit/compare/0.11.0...main)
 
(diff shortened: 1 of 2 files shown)


antonio-gg-dev (approved):

Chemaclass (approved):
Thanks!

Issue #257, opened 2024-06-15
use bashunit assertions outside of test scripts


staabm:
we are in the process of integrating bashunit into the PHPStan project.

see https://github.com/phpstan/phpstan-src/pull/3160 for the initial draft in which I try to convert a few existing tests into a bashunit

---

our primary goal is to make our existing tests more readable (for people not used to bash).

e.g. one of our current tests look like:

```
cd e2e/trait-caching
../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
patch -b data/TraitOne.php < TraitOne.patch
OUTPUT=$(../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/ || true)
echo "$OUTPUT"
[ $(echo "$OUTPUT" | wc -l) -eq 1 ]
grep 'Method TraitsCachingIssue\\TestClassUsingTrait::doBar() should return stdClass but returns Exception.' <<< "$OUTPUT"
```

two things we want to improve on this test:

1. `grep` is not readable.. atm one needs to be aware of chars to escape.. `assert_contains` is way better 

could we use the `bashunit` assertions outside a test? e.g. `bashunit assert_contains "my needle" "$OUTPUT"` ?
is such construction somehow supported? or maybe we can "import" bashunit assertion functions into the current bash, so we can just call them with `assert_contains`?

2. using a more readable variant of counting lines in $OUTPUT. is it possible to register custom assertion functions? or would a new core assertion function like `assert_line_count` be possible?

ondrejmirtes:
Basically we want to use the nice assertions syntax of bashunit not to unit tests bash functions, but to end-to-end tests executables.

Is this supported by bashunit? Do you have any tips for us?

I've been following the project for a long time and I'm a huge fan of your approach (release early and often)!

Chemaclass:
@staabm @ondrejmirtes I like the idea of implementing a feature to allow using bashunit functions isolated without the tests. Actually, you already prepared a very nice starting point here https://github.com/phpstan/phpstan-src/pull/3160/files#r1641646749 
I will work out something today focus on the DX integrated on the bashunit core itself

ondrejmirtes:
That would be awesome! Also it will make you cramp more content into your bashunit talk next time you give it I bet 😀

Chemaclass:
[...]

PR #258, opened 2024-06-16
Allow calling assertions outside tests

Chemaclass:
## 📚 Description

Issue: https://github.com/TypedDevs/bashunit/issues/257

## 🔖 Changes

- Add an option `-a|--assert` to `./bashunit` to allow running an [assertion](https://bashunit.typeddevs.com/assertions) from bashunit without the test context
- Benefits: provides the visual output from bashunit assertions + the result exit code is 0(passes) or 1(failures)

## ✅ To-do list

### DIFF
--- a/bashunit
+++ b/bashunit
@@ -17,6 +17,7 @@ source "$BASHUNIT_ROOT_DIR/src/helpers.sh"
 source "$BASHUNIT_ROOT_DIR/src/upgrade.sh"
 source "$BASHUNIT_ROOT_DIR/src/assertions.sh"
 source "$BASHUNIT_ROOT_DIR/src/runner.sh"
+source "$BASHUNIT_ROOT_DIR/src/main.sh"
 
 ###############
 #### MAIN #####
@@ -83,11 +84,6 @@ if [[ -n $_ASSERT_FN ]]; then
   if [[ "$(state::get_tests_failed)" -gt 0 ]] || [[ "$(state::get_assertions_failed)" -gt 0 ]]; then
       exit 1
   fi
-
 else
-  console_header::print_version_with_env
-  runner::load_test_files "$_FILTER" "${_FILES[@]}"
-  console_results::render_result
-  exit 0
+  main::exec_tests "$_FILTER" "${_FILES[@]}"
 fi
-
--- a/bashunit
+++ b/bashunit
@@ -22,12 +22,18 @@ source "$BASHUNIT_ROOT_DIR/src/runner.sh"
 #### MAIN #####
 ###############
 
+_ASSERT_FN=""
 _FILTER=""
 _FILES=()
 
 while [[ $# -gt 0 ]]; do
   argument="$1"
   case $argument in
+    -a|--assert)
+      _ASSERT_FN="$2"
+      shift
+      shift
+      ;;
     -f|--filter)
       _FILTER="$2"
       shift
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Awesome!

tests/acceptance/bashunit_direct_fn_call_test.sh
```diff
@@ -0,0 +1,26 @@
...
+  ./bashunit -a assert_equals --env "$TEST_ENV_FILE" "$expected" $actual
+  assert_successful_code
+}
+
+function test_bashunit_direct_fn_call_failure() {
+  local expected="foo"
+  local actual="bar"
+
+  assert_match_snapshot "$(./bashunit -a assert_equals --env "$TEST_ENV_FILE" "$expected" $actual)"
+  assert_general_error "$(./bashunit -a assert_equals --env "$TEST_ENV_FILE" "$expected" $actual)"
```
staabm, line 20: since we already use a `-a` or `--assert` as cli option, wdyt about leaving the `assert_` prefix out of the argument...?

would be less to write and less duplication

```suggestion
  assert_match_snapshot "$(./bashunit -a equals --env "$TEST_ENV_FILE" "$expected" $actual)"
  assert_general_error "$(./bashunit -a equals --env "$TEST_ENV_FILE" "$expected" $actual)"
```

Chemaclass: What about allowing both? It cost nothing really, and some developers might be purist in the sense they want to use the complete actual assert function (with the prefix), and some (like you) might prefer without `assert_` prefix to avoid "duplication". Implementing this is pretty easy, just when checking if the function exists, in case it does not, then try again prepending the prefix

antonio-gg-dev (approved):

src/default_env_config.sh
```diff
@@ -8,4 +8,5 @@ _DEFAULT_SIMPLE_OUTPUT=false
 _DEFAULT_STOP_ON_FAILURE=false
 _DEFAULT_SHOW_EXECUTION_TIME=true
 _DEFAULT_DEFAULT_PATH=
+_DEFAULT_ASSERT_FN=""
```
Chemaclass, line 11: This is not needed

khru (approved):

bashunit
```diff
@@ -17,17 +17,20 @@ source "$BASHUNIT_ROOT_DIR/src/helpers.sh"
 source "$BASHUNIT_ROOT_DIR/src/upgrade.sh"
 source "$BASHUNIT_ROOT_DIR/src/assertions.sh"
 source "$BASHUNIT_ROOT_DIR/src/runner.sh"
+source "$BASHUNIT_ROOT_DIR/src/main.sh"
 
-###############
-#### MAIN #####
-###############
-
+_ASSERT_FN=$_DEFAULT_ASSERT_FN
```
Chemaclass, line 22: Todo: rollback to empty str directly here. This is not a value from the .env but an option passed via the command line only and it should be always empty by default.

khru: It's not, but it could be and use a specific .env for CI

khru: It's not, but it could be and use a specific .env for CI

tests/acceptance/bashunit_direct_fn_call_test.sh
staabm: Having both would also be great, thanks

docs/standalone.md
```diff
@@ -0,0 +1,40 @@
+# Standalone
+
+You can use all bashunit assertions outside any tests if you would like to use them in your already integration or e2e tests, for entire applications or executables.
```
staabm, line 3:
```suggestion
You can use all bashunit assertions outside any tests if you would like to use them in integration or end-to-end tests, for entire applications or executables.
```

staabm (approved):
its awesome, thank you.

JesusValera (approved):

bashunit
Chemaclass: This is something that you use with `./bashunit -a|--assert`. I don't see how it "could be used for CI" from the `.env`? Or am I missing something? Could you elaborate a bit more on what's the value on having this here, @khru?

Chemaclass: After talking with @khru and @antonio-gg-dev we agreed on this, so I will revert it 👍🏼

PR #259, opened 2024-06-17
Add `assert_line_count`

staabm:
## 📚 Description

Implement new `assert_line_count` assertion function.

refs https://github.com/TypedDevs/bashunit/issues/257

## 🔖 Changes

- Add `assert_line_count`

## ✅ To-do list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Add multi-invokers; consolidate parameterized-testing documentation
 - Add `fail()` function
 - Remove all test mocks after each test case
+- Add `assert_line_count`
 
 ## [0.11.0](https://github.com/TypedDevs/bashunit/compare/0.10.1...0.11.0) - 2024-03-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased](https://github.com/TypedDevs/bashunit/compare/0.12.0...main)
 
 - Allow calling assertions standalone outside tests
+- Add `assert_line_count`
 
 ## [0.12.0](https://github.com/TypedDevs/bashunit/compare/0.11.0...0.12.0) - 2024-06-11
 
@@ -11,7 +12,6 @@
 - Add multi-invokers; consolidate parameterized-testing documentation
 - Add `fail()` function
 - Remove all test mocks after each test case
-- Add `assert_line_count`
 
 ## [0.11.0](https://github.com/TypedDevs/bashunit/compare/0.10.1...0.11.0) - 2024-03-02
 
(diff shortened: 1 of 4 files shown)


src/assert.sh
```diff
@@ -315,3 +315,23 @@ function assert_greater_or_equal_than() {
...
+
+function assert_line_count() {
+  local expected="$1"
+  local actual="$2"
+  local label="${3:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
+
+  if [ -z "$actual" ]; then
+    local actual_line_count=0
+  else
+    local actual_line_count=$(echo "$actual" | wc -l | tr -d '[:blank:]')
```
staabm, line 327: on my macos, `wc -l` returns a number prepended by whitespaces. therefore I used `tr` to strip them

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Add multi-invokers; consolidate parameterized-testing documentation
 - Add `fail()` function
 - Remove all test mocks after each test case
+- Add `assert_line_count`
```
Chemaclass, line 14: Can you please move this up to the Unreleased section? The `0.12` was already released https://github.com/TypedDevs/bashunit/releases/tag/0.12.0

docs/assertions.md
```diff
@@ -153,6 +153,27 @@ function test_failure() {
...
+Reports an error if `haystack` does not contain `count` lines.
+
+::: code-group
+```bash [Example]
+function test_success() {
+  local string="this is line one
+this is line two
+this is line three"
+
+  assert_line_count "3" "$string"
```
Chemaclass, line 168: Can we use literal numbers in the example instead of strings? I assume it would work, and it makes more sense. What do you think? (Same for line 172 here, and in the unit tests)
```suggestion
  assert_line_count 3 "$string"
```

Chemaclass (commented):
Very cool! Just some minor comments

CHANGELOG.md
staabm: I tried to sneak this in with my wayback machine. didn't work out it seems ;)

docs/assertions.md
staabm: fixed, thanks

staabm: btw: just noticed in other examples - e.g. `assert_less_than` - we use quotes arround numbers

Chemaclass: I guess for bash is the same (everything is a string...), and probably there is a mix between the old and new codebase. That said, we would like to keep a bit `(strict_types=1)` mindset in the project from now on. Feel free to submit a refactoring PR for this if you see any opportunity! :)

Chemaclass (approved):
Thank you!

PR #264, opened 2024-06-17
Fix: assert_line_count

Chemaclass:
## 📚 Description

Related to: https://github.com/phpstan/phpstan-src/pull/3160#discussion_r1643031961 
> The `\n` are ignored when counting the lines `actual=$(echo "$input_str" | wc -l | tr -d '[:blank:]')`

## 🔖 Changes

- Count additionally the `\n` of a string on `assert_line_count`

### DIFF
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -318,21 +318,25 @@ function assert_greater_or_equal_than() {
 
 function assert_line_count() {
   local expected="$1"
-  local actual="$2"
+  local input_str="$2"
   local label="${3:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
 
-  if [ -z "$actual" ]; then
-    local actual_line_count=0
+  if [ -z "$input_str" ]; then
+    local actual=0
   else
-    local actual_line_count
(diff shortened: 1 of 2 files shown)


src/assert.sh
```diff
@@ -318,21 +318,23 @@ function assert_greater_or_equal_than() {
...
-  if [ -z "$actual" ]; then
-    local actual_line_count=0
+  if [ -z "$input_str" ]; then
+    local actual=0
   else
-    local actual_line_count
-    actual_line_count=$(echo "$actual" | wc -l | tr -d '[:blank:]')
+    local actual
+    actual=$(echo "$input_str" | wc -l | tr -d '[:blank:]')
+    additional_new_lines=$(grep -o '\\n' <<< "$input_str" | wc -l | tr -d '[:blank:]')
```
staabm, line 329: I don't get why/how bash represents the newline in 2 different characters.. but the tests seem to show it works.

I don't know why 😅

Chemaclass: Same here... I just got to this solution driven by the tests, I don't understand why this works... so it's bash... 🤣 
> And here, yet, another proof that TDD works 😝

fabriziofs (approved):

PR #265, opened 2024-06-18
fix: assert_line_count on multiline

Chemaclass:
## 📚 Description

Issue: https://github.com/phpstan/phpstan-src/pull/3160#discussion_r1643902055

Multiline strings are passes as multiple args to `assert_line_count` which makes it complicated to pass around and manipulate as unique string.

## 🔖 Changes

- `assert_line_count` uses first arg for the expected and the rest (think about a `variadic ...string arg`) will be the string itself that will be use to count its lines.

 ## 💡 Extra
Tradeoff: I had to rollback https://github.com/TypedDevs/bashunit/pull/262 in order to make it working. I think I can make it back again, but it might take another couple of hours... so, I thought to ship this first, make it work, and then in a follow up make it better.

### DIFF
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -318,8 +318,9 @@ function assert_greater_or_equal_than() {
 
 function assert_line_count() {
   local expected="$1"
-  local input_str="$2"
-  local label="${3:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
+  local input_arr=("${@:2}")
+  local input_str
+  input_str=$(printf '%s\n' "${input_arr[@]}")
 
   if [ -z "$input_str" ]; then
     local actual=0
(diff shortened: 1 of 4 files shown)


src/assert.sh
```diff
@@ -318,8 +318,9 @@ function assert_greater_or_equal_than() {
 
 function assert_line_count() {
```
staabm, line 319: I feel sorry for opening this can of worms :-).

thanks for working thru all these problem.

Chemaclass: You opened a pandora box... nah, it is also fun - even though I really dislike bash 😝

staabm: "you just need to count the lines" - they said

tests/acceptance/bashunit_direct_fn_call_test.sh
```diff
@@ -28,6 +28,15 @@ second line"
...
 
+function test_bashunit_assert_line_count_with_explicit_line_break() {
+  local actual="first line
+second line
+\n four line"
+
+  ./bashunit -a line_count 4 "$actual"
+  assert_successful_code
+}
+
```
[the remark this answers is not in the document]
Chemaclass: I would prefer merging it with the other one, to have just one like 2e2 test. Can you do that directly in the PR, @fabriziofs ?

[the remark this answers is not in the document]
fabriziofs: Done 👌

fabriziofs (approved):

PR #266, opened 2024-06-18
Fix: multistring assertions

Chemaclass:
## 📚 Description

Related: https://github.com/phpstan/phpstan-src/actions/runs/9567238902/job/26374488151 
> Thanks for the support and help @staabm

## 🔖 Changes

- Remove the optional label and use always `label="$(helper::normalize_test_function_name "${FUNCNAME[1]}")"`, which is something we are using all around and it makes sense to keep it this way
- For string arguments, use all arguments but the first one (reserved for the expected)

## ⏰  Disclaimer 

- `assert_equals` on multiline strings does not work, because multilines takes mutiple arguments positions when passing to other functions... so it's bash - at least I didn't find another solution for this problem so far.
- This means that, multiline strings will work as last argument as they are treated as variadic arguments like `...string`, therefore the first argument must be `one liner`, whatever that is.
> You can use `assert_contains` or `assert_matches` instead

### DIFF
--- a/src/assert.sh
+++ b/src/assert.sh
@@ -2,8 +2,9 @@
 
 function fail() {
   local message=$1
-  local label="${2:-$(helper::normalize_test_function_name "${FUNCNAME[1]}")}"
 
+  local label
+  label="$(helper::normalize_test_function_name "${FUNCNAME[1]}")"
   state::add_assertions_failed
   console_results::print_failure_message "${label}" "$message"
 }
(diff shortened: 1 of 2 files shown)


staabm:
> assert_equals on multiline strings does not work, because multilines takes mutiple arguments positions when passing to other functions... so it's bash - at least I didn't find another solution for this problem so far.


Would it make sense to emit an error in case of multi-line strings telling the user that its not supported?

JesusValera (approved):

staabm:
I see, thanks for the details. maybe it makes sense to open a new issue about this known problem, so people showing up in the issue tracker find the "known problem"..?

PR #296, opened 2024-07-15
Add versioning to the webpage

JesusValera:
## 📚 Description

Implements multiple versions to bashunit.

This will help the users to see which specific functionalities are available on the specific version they are using.

I am using the following external library: https://github.com/IMB11/vitepress-versioning-plugin/tree/master
Whose documentation is: https://vvp.imb11.dev/

## 🔖 Changes

Add a new button on top to let the users pick the desired version

image

## Caveats

- [ ] Show the elements in the sidebar (not sure how to do it at the moment)
- [ ] Store the current version in a variable, then, when we click on the logo, we will go to `home/{version}`
  - right now, when you press the logo, you go to `home` independently the version you were before
- [ ] If you typed some URL (eg: `home/asserts`), redirect to the latest version (eg: `home/0.14.0/asserts`)
- [ ] Mark the currently selected version as `green` in the dropdown when we are not on the homepage (eg: if we go to `/x.xx.xx/asserts`, the version is not selected anymore)
image

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -1,14 +1,26 @@
-import { defineConfig } from 'vitepress'
+import { DefaultTheme } from "vitepress";
+import defineVersionedConfig from 'vitepress-versioning-plugin';
 import pkg from '../../package.json'
 
 // https://vitepress.dev/reference/site-config
-export default defineConfig({
+export default defineVersionedConfig({
   title: 'bashunit - A simple testing library for bash scripts',
   titleTemplate: 'bashunit',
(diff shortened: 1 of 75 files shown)


Chemaclass:
How valid/up-to-date is this PR, @JesusValera ?

JesusValera:
> How valid/up-to-date is this PR, @JesusValera ?

This PR is not valid anymore.
@antonio-gg-dev mentioned it was possible to get the same functionality using Cloudflare and cache, which was much simpler and cleaner.

I will close this PR then.

PR #304, opened 2024-08-11
Remove ProductHunt & Manfred Awards from index

JesusValera:
## 📚 Description

Remove banners from index's bottom page

image


Chemaclass (approved):
Yeah, good point. I think it doesnt align well with the current style of the website. We should find another way or section where to put these things

Issue #322, opened 2024-09-12
assert files are equal


staabm:
I wonder whether there is a assertion which allows me to compare a file which contains the expected content with one containing the actual content?

Chemaclass:
We don't have that right now, but it should be easy to include! Very nice idea.
If you want to do it: check the existing https://bashunit.typeddevs.com/assertions#assert-is-file works, and you can implement a similar for assert_file_equals or similar. Otherwise, I can work on it this weekend

PR #326, opened 2024-09-13
Added assert_files_equals

Chemaclass:
## 📚 Description

Closes: https://github.com/TypedDevs/bashunit/issues/322 by @staabm

## 🔖 Changes

- Add `assert_files_equals`
- Add `assert_files_not_equals`
- Fixed "unbound variable" on runtime error message

## 🎥  Demo

https://github.com/user-attachments/assets/31828c38-e676-48ed-9729-12ddc43038eb

## ✅ To-do list 

### DIFF
--- a/src/assert_files.sh
+++ b/src/assert_files.sh
@@ -56,12 +56,14 @@ function assert_files_equals() {
   local expected="$1"
   local actual="$2"
 
-  if ! cmp -s "$expected" "$actual"; then
+  if ! diff -u "$expected" "$actual"; then
     local label
     label="$(helper::normalize_test_function_name "${FUNCNAME[1]}")"
     state::add_assertions_failed
-    console_results::print_failed_test "${label}" "${expected}" "compared" "${actual}" \
-        "Diff" "$(diff -u "$expected" "$actual" || true)"
+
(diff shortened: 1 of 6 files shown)


src/assert_files.sh
```diff
@@ -51,3 +51,19 @@ function assert_is_file_empty() {
...
+function assert_files_equals() {
+  local expected="$1"
+  local actual="$2"
+
+  if ! cmp -s "$expected" "$actual"; then
+    local label
+    label="$(helper::normalize_test_function_name "${FUNCNAME[1]}")"
+    state::add_assertions_failed
+    console_results::print_failed_test "${label}" "${expected}" "compared" "${actual}" \
+        "Diff" "$(diff -u "$expected" "$actual" || true)"
```
staabm, line 64: thanks for working on my request <3.

does the diff command already return a exit code which can be used at the IF statement?
I think the results would be more consistent, if we don't use 2 different command to compare the files and to print the diff

Chemaclass: Very good point! I will change it 🧠

antonio-gg-dev (approved):

PR #331, opened 2024-09-16
Add support for Alpine (Linux Distro)

skinner-m-c:
## 📚 Description

Add automation for running Bashunit tests in Alpine locally and with Github actions.

## 🔖 Changes

Adds a target in the Makefile for running tests in Alpine. There are other approaches that could be done (e.g. entirely in Github actions, but that would make contributors unable to test changes in Alpine locally without additional work).

This change does not include fixes for running tests in Alpine. I do have some fixes that I was going to do in a separate pull request because they are a bit involved. 

There are several ways to implement this kind of testing. Let me know if there is a preferred approach.

This automation does fail because `bashunit` does not currently pass in Alpine. Because this change will cause builds to fail, I recommend commenting or excluding out the Github action changes for running these tests and uncomment or add them back when the Alpine test pass. I have included the Github action for sake of completion. 

## ✅ To-do list


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added `assert_files_equals`, `assert_files_not_equals`
 - Fixed total tests wrong number
 - Added `BASHUNIT_TESTS_ENV`
+- Added support for running tests in Alpine
```
Chemaclass, line 13: Please, do a git pull to fetch the latest changes as the 0.16 was released this weekend, so you can add this line in a new section on top as `## Unreleased`

Makefile
```diff
@@ -67,6 +67,16 @@ test/watch: $(TEST_SCRIPTS)
 	@./bashunit $(TEST_SCRIPTS)
 	@fswatch -m poll_monitor -or $(SRC_SCRIPTS_DIR) $(TEST_SCRIPTS_DIR) .env Makefile | xargs -n1 ./bashunit $(TEST_SCRIPTS)
 
+test-alpine:
+	# Run tests for Alpine in a container with only the very basic dependencies. Bashunit should work for the base
+	# Alpine installation.
+	docker run -v $$(pwd):/working --rm alpine:3.20 /bin/sh -c " \
```
khru, line 73: The alpine version should be on a .env file

skinner-m-c:
Thanks for approving the test runs. It has now reached the first error for Alpine that I intend to fix in other pull request. Let me know what you think should be done next. I can remove the Github action and re-add it at a later time when Alpine test pass.

.github/CONTRIBUTING.md
```diff
@@ -55,6 +55,7 @@ Run tests from the library:
 ```bash
 # using make
 make test
+make alpine-test
```
Chemaclass, line 58: I started checking this deeper, and this is the first thing that I found so far hehe

```suggestion
make test-alpine
```

Chemaclass:
@skinner-m-c I am new with Alpine, so I might be missing something here... I wanted to checkout the branch and play around with it, but I got stuck on the very first step of installing the img. 

First, I tried with the make command:
Screenshot 2024-09-21 at 20 59 41

I also tried with pure raw docker:
Screenshot 2024-09-21 at 20 58 07

I don't get how to pass around this error 🤔  Any ideas, @skinner-m-c ?
```
chown: /working/.git/objects/61/f510da657127cd9580655326b33463adafd810: Permission denied
```

Chemaclass:
Update: I managed to reproduce it 
Screenshot 2024-09-21 at 22 06 15

```bash
docker run --rm -it \
  -v "$(pwd)":/project \
  -w /project \
  alpine:3.20 \
  sh -c "apk add bash make && bash"
# and then running bashunit from the inside
./bashunit
```

Chemaclass:
@skinner-m-c, please double-check the changes and let me know your thoughts. 
As soon as you are okay with them, we can merge them.

tests/unit/directory_test.sh
```diff
@@ -127,7 +127,7 @@ function test_unsuccessful_assert_is_directory_readable_when_a_file_is_given() {
 }
 
 function test_unsuccessful_assert_is_directory_readable_without_execution_permission() {
-  if [[ "$_OS" == "Windows" ]]; then
+  if [[ "$_OS" == "Windows" || $_DISTRO = "Alpine" ]]; then
```
Chemaclass, line 130: I got these `tests/unit/directory_test.sh` failing on Alpine for some reason (directory permissions), but I don't know how to solve them now, so I'd better skip them for now, and work on them in another iteration/PR
```bash
docker run --rm -it -v "$(pwd)":/project -w /project alpine:latest \
    sh -c  "apk add bash make shellcheck git curl perl && ./bashunit tests/unit/directory_test.sh"
```
Screenshot 2024-09-22 at 14 57 09

Chemaclass (approved):

Issue #305, opened 2024-08-20
Data provider are too slow


nosrio:
| Q                | A                       |
|------------------|-------------------------|
| OS               | Ubuntu 20.04.4 LTS |
| Shell            | bash          |
| bashunit version | 0.14.0                 |

#### Summary

Using data provider to interate on long list decrease performance of tests.

#### Current behavior

If I use this code:

```
function test_for_each(){
    local elements=$(seq 1 1 100)
    local regex='^[[:digit:]]+$'

    for element in ${elements[@]}
    do
        assert_matches "${regex}" "${element}"
    done
}
```
The output is 

```
bashunit - 0.14.0
Running tests/parallel_test.sh
✓ Passed: For each

Tests:      1 passed, 1 total
Assertions: 100 passed, 100 total

 All tests passed 
Time taken: 149 ms
```

However if I change the code and use data provider:

```
function provider_elements(){
    local elements=$(seq 1 1 100)
    echo ${elements[@]}
}

# data_provider provider_elements
function test_for_each(){
    local regex='^[[:digit:]]+$'
    local element=${1}

    assert_matches "${regex}" "${element}"
}
```

The output is:
```
bashunit - 0.14.0
Running tests/parallel_test.sh
✓ Passed: For each (1)
✓ Passed: For each (2)
✓ Passed: For each (3)
.....
✓ Passed: For each (100)

Tests:      100 passed, 100 total
Assertions: 100 passed, 100 total

 All tests passed 
Time taken: 3706 ms
```

This is 24x time slower and it get worse when N is greater.

#### How to reproduce

See code above.
I run strace on both examples, maybe this would help:

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 77,65    0,080395         176       455       216 wait4
  9,47    0,009803          41       239           clone
  2,11    0,002184           1      2110           rt_sigprocmask
  1,70    0,001757           2       771           read
  1,41    0,001458           1      1428        46 close
  1,06    0,001101           0      3152           rt_sigaction
  0,99    0,001021           1       921           mmap
  0,94    0,000973          24        40           execve
  0,91    0,000945           4       216           pipe
  0,74    0,000771           1       482       138 openat
  0,66    0,000688           0      1347      1048 stat
  0,49    0,000505           1       257           mprotect
  0,31    0,000319           1       216           rt_sigreturn
  0,24    0,000245           1       180           brk
[...]

PR #332, opened 2024-09-21
Improve debug output

Chemaclass:
## 📚 Description

Inspired by wanting debug this in an easier way https://github.com/TypedDevs/bashunit/issues/305#issuecomment-2365045807 @staabm 

## 🔖 Changes

- Add optional file-path as 2nd arg to `--debug` option

### BEFORE

```bash
./bashunit local/example_test.sh --debug > local/debug.sh 2>&1
```

### AFTER
```bash
./bashunit local/example_test.sh --debug local/debug.sh
```

## ✅ To-do list

### DIFF
--- a/bashunit
+++ b/bashunit
@@ -47,6 +47,10 @@ while [[ $# -gt 0 ]]; do
       export BASHUNIT_SIMPLE_OUTPUT=false
       ;;
     --debug)
+      OUTPUT_FILE="${2:-}"
+      if [[ -n $OUTPUT_FILE ]]; then
+        exec > "$OUTPUT_FILE" 2>&1
+      fi
       set -x
       ;;
     -S|--stop-on-failure)
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ bin/checksum
 # logger
 log-junit.xml
 report.html
+
+# internal
+local/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add optional file-path as 2nd arg to `--debug` option
 - Rename `BASHUNIT_TESTS_ENV` to `BASHUNIT_LOAD_FILE`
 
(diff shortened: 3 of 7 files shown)


bashunit
```diff
@@ -47,6 +47,10 @@ while [[ $# -gt 0 ]]; do
       export BASHUNIT_SIMPLE_OUTPUT=false
       ;;
     --debug)
+      OUTPUT_FILE="${2:-}"
+      if [[ -n $OUTPUT_FILE ]]; then
+        exec > "$OUTPUT_FILE" 2>&1
```
staabm, line 52: Why do you need this option? Isn't it sufficient to redirect the output directly in the cli command?

`bashunit --debug > my.log` ?

Chemaclass: Nope, in that case the redirection is the result output. The --debug enable `set -x` which produces a runtime output that is not redirected. That is why you need to change the FD adding `2>&1` to make it work as expected. For this reason, I thought simplifying this to make it work as expected
```bash
./bashunit local/example_test.sh --debug > local/debug.sh 2>&1
```

staabm: Is it the same as `./bashunit local/example_test.sh --debug 2> local/debug.sh`

?

Chemaclass: I thought so at the beginning. The main diff:
- `./bashunit local/example_test.sh --debug 2> local/debug.sh`
Captures only error messages in the file while allowing regular output to go to the terminal.

- `./bashunit local/example_test.sh --debug > local/debug.sh 2>&1`
Captures both output and error messages into the same file.

---

However, for some reason that I still dont know, when using `./bashunit tests/acceptance --debug 2> local/debug.sh` I get an error on runtime (it might need investigation why)
Screenshot 2024-09-21 at 12 21 10
Screenshot 2024-09-21 at 12 22 02

while when doing it with `./bashunit tests/acceptance --debug > local/debug.sh 2>&1` I dont get any error 
Screenshot 2024-09-21 at 12 24 02
Screenshot 2024-09-21 at 12 24 14

--- 

My guess this error is due to how the FD2 (for errors) is handled... I am still having struggle understanding these bash FDs 😄

JesusValera (approved):

PR #333, opened 2024-09-21
Add test duration

Chemaclass:
## 📚 Description

Right now, you can see total exec time only at the end but not on the individual test level.

## 🔖 Changes

Display test runtime if `BASHUNIT_SHOW_EXECUTION_TIME` is enabled

## ✅ To-do list

### BEFORE

Screenshot 2024-09-21 at 18 03 38
Screenshot 2024-09-21 at 18 03 58

### AFTER

Screenshot 2024-09-21 at 18 01 26
Screenshot 2024-09-21 at 18 03 01


src/console_header.sh
```diff
@@ -37,15 +37,15 @@ EOF
     if [ "$total_tests" -eq 0 ]; then
       printf "%s\n" "$BASHUNIT_VERSION"
     else
-      printf "%s | Total tests: %s\n" "$BASHUNIT_VERSION" "$total_tests"
+      printf "%s | Tests: ~%s\n" "$BASHUNIT_VERSION" "$total_tests"
```
Chemaclass, line 40: Reason: counting total tests is easy, but not the data providers. Therefore, I would like an overview of the number of tests. Still, I don't mind (at least for now) the exact consideration of data_provider level, especially considering that this might be performance-wise costly... so I would rather postpone this (if even needed). 

Therefore, I added the ~ to clarify that this is an honest estimation.

JesusValera (approved):

Issue #334, opened 2024-09-22
Make asserting failed exit codes and output nicer


ondrejmirtes:
Creating this based on our discussion in Dresden :)

If we want to assert that a command failed & assert its output, we have to do something like this:

```
OUTPUT=$(../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/ || true)
echo "$OUTPUT"
../bashunit -a line_count 1 "$OUTPUT"
../bashunit -a contains 'Method TraitsCachingIssue\TestClassUsingTrait::doBar() should return stdClass but returns Exception.' "$OUTPUT"
```

The problem is that we're not really asserting that the commands returned exit code greater than zero, and also that we need to do the whole `OUTPUT=$(command || true)` dance.

Would be great if Bashunit provided a nicer way to do this :)

Chemaclass:
@ondrejmirtes Can you give me an example of how would you vision the ideal use case? Especially, how would you use **bashunit** in this context (what would you like to have). This can help me going in the direction you need (test-first mindset), because right now I am not sure if I got the solution/problem right.
Also, if you can give me a link of a real example of this in your CI that might help too (to contextualize even more)

Chemaclass:
I am playing around trying to feel the pain/problem
```php
<?php
class Testing {
  function bar(int $baz = 0) {
    return $baz;
  }
}
```
```bash
# Current example
OUTPUT=$("$PHPSTAN_PATH" analyze --no-progress --level 8 --error-format raw local/ || true)
echo "$OUTPUT"
./bashunit -a line_count 1 "$OUTPUT"
./bashunit -a contains 'Method teseting::bar() has no return type specified.' "$OUTPUT"

# Alternative(?)
OUTPUT=$("$PHPSTAN_PATH" analyze --no-progress --level 8 --error-format raw local/)
./bashunit -a line_count 1 "$OUTPUT"
./bashunit -a contains 'Method teseting::bar() has no return type specified.' "$OUTPUT"
```

Both example works just fine. This is why I mention I feel I need more context, because using the `Alternative()?` feels right to me, so most luckily I am missing something here :)

staabm:
there is still no assertion for the exit code of the phpstan command

Chemaclass:
@staabm you can assert the exit codes as well: https://bashunit.typeddevs.com/assertions#assert-exit-code

staabm:
The github action will stol executing further commands as soon as a commands exits with non-zero. 

[...]

PR #336, opened 2024-09-24
Make asserting failed exit codes and output nicer

Chemaclass:
## 📚 Description

Closes: https://github.com/TypedDevs/bashunit/issues/334

## 🔖 Changes

- Allow defer calling a function when asserting `exit_code` by passing the callable command to evaluate as raw string instead of `$(...)`
- bashunit output will be render on the stderr

## 🖼️  Demo

Render the stdout in the terminal and the errors in a file

Screenshot 2024-09-28 at 10 21 54

```bash
export PHPSTAN_PATH=[...]/vendor/bin/phpstan

../bashunit -a exit_code "1" "$PHPSTAN_PATH analyze --no-progress --level 8 --error-format raw ./"
echo $?

../bashunit -a exit_code "0" "$PHPSTAN_PATH analyze --no-progress --level 8 --error-format raw ./"
echo $?

../bashunit -a exit_code "0" "$PHPSTAN_PATH analyze --no-progress --level 8 --error-format raw ./" 2> /tmp/error.log
cat /tmp/error.log
```

## ✅ To-do list

Screenshot 2024-09-28 at 10 25 43

### DIFF
--- a/src/main.sh
+++ b/src/main.sh
@@ -34,8 +34,7 @@ function main::exec_tests() {
 
 function main::exec_assert() {
   local original_assert_fn=$1
-  local forward_stdout=$2
-  local args=("${@:3}")
+  local args=("${@:2}")
 
   local assert_fn=$original_assert_fn
 
@@ -56,7 +55,7 @@ function main::exec_assert() {
   local bashunit_exit_code=0
   local output
 
-  if [[ "$last_arg" == eval* ]]; then
+  if [[ "$assert_fn" == "assert_exit_code" && "$last_arg" == eval* ]]; then
     local callable_command="${last_arg#eval }"
     output=$(eval "$callable_command" 2>&1 || echo "inner_exit_code:$?")
 
@@ -74,8 +73,8 @@ function main::exec_assert() {
   fi
 
   if [[ -n "$output" ]]; then
-    [[ "$forward_stdout" == "true" ]] && echo "$output" 1>&1
-    [[ "$assert_fn" == "assert_exit_code" ]] && assert_fn="assert_same"
+    echo "$output" 1>&1
+    assert_fn="assert_same"
   fi
 
   # Run the assertion function and write into stderr
--- a/src/main.sh
+++ b/src/main.sh
@@ -42,7 +42,6 @@ function main::exec_assert() {
   if ! type "$assert_fn" > /dev/null 2>&1; then
     assert_fn="assert_$assert_fn"
     if ! type "$assert_fn" > /dev/null 2>&1; then
-      # Write into stderr
       echo "Function $original_assert_fn does not exist." 1>&2
       exit 127
     fi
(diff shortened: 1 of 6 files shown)


src/main.sh
```diff
@@ -34,21 +34,56 @@ function main::exec_tests() {
...
+  # Get the last argument safely by calculating the array length
+  local last_index=$((${#args[@]} - 1))
+  local last_arg="${args[$last_index]}"
+  local inner_exit_code=0
+  local bashunit_exit_code=0
+  local output
+
+  if [[ "$assert_fn" == "assert_exit_code" && "$last_arg" == eval* ]]; then
+    local callable_command="${last_arg#eval }"
+    output=$(eval "$callable_command" 2>&1 || echo "inner_exit_code:$?")
```
Chemaclass, line 60: `eval` is a real bash command, but in this context of last argument I am using it at `bashunit` level so hide the complexity and bring familiarity with the "bash context" without bringing new weird terms. This is the trick:
```bash
# remove the "eval" word from the last argument
local callable_command="${last_arg#eval }"

# use real bash "eval" to evaluate your command
output=$(eval "$callable_command" 2>&1 || echo "inner_exit_code:$?")
```

src/main.sh
```diff
@@ -34,21 +34,61 @@ function main::exec_tests() {
...
+  local last_arg="${args[$last_index]}"
+  local inner_exit_code=0
+  local bashunit_exit_code=0
+  local output
+
+  if [[ "$assert_fn" == "assert_exit_code" && "$last_arg" == eval* ]]; then
+    # remove the "eval" word from the last argument
+    local callable_command="${last_arg#eval }"
+    # use real bash "eval" to evaluate the command
+    output=$(eval "$callable_command" 2>&1 || echo "inner_exit_code:$?")
```
Chemaclass, line 62: `eval` is a real bash command, but in this context of last argument I am using it at `bashunit` level to hide the complexity and bring "familiarity" without bringing new terms.

ondrejmirtes:
I have some ideas about this (which you don't have to follow but maybe just think about the direction):

1) If you want `bashunit -a exit_code "1" "eval the_command"` then what would `bashunit -a exit_code "1" "the_command"` mean? Does it mean anything? Can we remove the "eval" and still have it work?

Does -a exit_code imply the thing in double quotes is going to be run as command?

Also trying to think of some other combinations that either need to mean something, or they need to error.

What would this do?

`./bashunit -a assert_same "foo" "eval the_command"`

The 2nd argument can just be a normal string. I think special strings shouldn't switch the behaviour of the command.

My preferred syntax would be to have:

`bashunit -e "the_command"`

It'd mean that bashunit would run the command and then assert what we ask it to assert. Right now the only assertion would be exit code:

`bashunit -a exit_code "1" -e "the_command"`

Maybe other asserts could work on the output automatically?

`bashunit -a exit_code "1" -a contains "some error" -e "the_command"` (I don't know if I can currently pass multiple `-a` args to perform multiple assertions)

Also it'd be nice to save the output to a variable with another option for further asserts:

`bashunit -a exit_code "1" -e "the_command" -o $OUTPUT`

Chemaclass:
I like your proposal of removing eval from the second string passed to the exit_code:
- `bashunit -a exit_code "1"` - compare with the last exit code
- `bashunit -a exit_code "1" "the_command"` - run the command inside bashunit and compare the codes
- `bashunit -a exit_code "1" "$(the_command)"` - run the command outside and compare the codes

Basically, the change I would suggest/do is keep the same behaviour (no need to introduce -e or -o) and only for the exit_code, if the second argument is a string then treat it as a callable (which will run internally with native bash `eval`). How does this sounds like, @ondrejmirtes ?

Issue #354, opened 2024-10-06
Enable running tests in parallel


Chemaclass:
Investigate how could we enable running tests in parallel.

### Tech hint

Multithreading is not possible in bash, but...

You can run a process in the background with `&` and then make the main process wait for all its children with `wait`.

Consider storing each test result in a temp-file (within a temp-folder for each test script), and then at the end read the result of each test to know the totals of the test execution.

PR #358, opened 2024-10-09
Running in parallel

Chemaclass:
## 📚 Description

Closes: https://github.com/TypedDevs/bashunit/issues/354

## 🔖 Changes

- Add parallel running: `-p|--parallel`
- Add possibility of disabling parallel execution with `--no-parallel`

## ✅ To-do list

### Running the entire test suite sync (~43s)

Screenshot 2024-10-10 at 01 12 16

### Running the entire test suite in parallel (~15s) - 65.12% improment! 🚀 

Screenshot 2024-10-10 at 01 39 12

### Running parallel tests verbose

Screenshot 2024-10-10 at 01 55 41

### Documentation

Screenshot 2024-10-10 at 00 34 07


CosmeValera (approved):
Good job as always !!!

khru (approved):

khru (commented):
I would just say that for me parallel testing should always be the default way to run, as it is not just for performance, but also to know that your tests do not depend on each other,

I would rather suggest that it be always active by default in subsequent versions and that it can be disabled. 

That's my opinion, I think that the value that active brings is always greater, and that only in case you need to disable it because your code needs it, you do it. 


This opens up other issues such as being able to prevent a test file from being launched in parallel, this is common in other testing frameworks.

khru:
The tests are not good in alpine

Chemaclass:
I would rather not keep the parallel testing as default behavior but as it was until now. 
This is the default behavior on PHPUnit, Pest, JUnit, among others. However I found that Jest run them in parallel "when it make sense" based on some heuristic -size of tests, number of tests, etc... but I find this too complicated to get into this right now. 

Aiming for simplicity, I am in favor of:
If you want parallel tests you can enable with `-p|--parallel` or in your .env `BASHUNIT_PARALLEL_RUN=true`

PR #376, opened 2024-10-13
Improve render file header on verbose

Chemaclass:
## 🔖 Changes

- Add the `Running $script` text always when verbose is enabled
  - Make it bold 

### No verbose

Screenshot 2024-10-13 at 13 48 52

#### Simple
Screenshot 2024-10-13 at 13 49 32

#### Detailed
Screenshot 2024-10-13 at 13 49 45

### DIFF
--- a/tests/acceptance/bashunit_execution_error_test.sh
+++ b/tests/acceptance/bashunit_execution_error_test.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2155
 set -euo pipefail
 
 function set_up_before_script() {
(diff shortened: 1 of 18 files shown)


tests/acceptance/bashunit_execution_error_test.sh
```diff
@@ -35,25 +21,24 @@ function test_bashunit_when_a_execution_error() {
...
+    printf "%s✗ Error%s: Error\n" "${color_red}" "${color_default}"
+    printf "    %sline 4: invalid_function_name: command not found%s\n" "${color_dim}" "${color_default}"
   )
-  fixture_end=$(
+  local fixture_end=$(
     format_summary_title "Tests:     "
     format_summary_value "1 failed" ", 1 total"
     format_summary_title "Assertions:"
-    format_summary_value "1 failed" ", 1 total"
+    format_summary_value "0 failed" ", 0 total"
```
Chemaclass, line 34: This is intentional, because a `command not found`. Context:
```bash
function test_error() {
  # before we had this as a way to find out that a command did not exist
  invalid_function_name
  assert_general_error

  # now bashunit is clever enough to know that a command does not exist by itself, 
  # without the need of an assertion afterwards
  invalid_function_name arg1 arg2
}
```

For this reason, there are no assertions in this test. Just a failing test.

CosmeValera (approved):

JesusValera (approved):

PR #380, opened 2024-10-17
Test that covers the source of a file and check the spies

khru:
## 🔖 Changes

- add test that covers the source of a file and check that the spies work


tmp/temp.sh
Chemaclass, line 1: do we want this tmp file? 🤔

Chemaclass (approved):

PR #392, opened 2025-02-12
fix: improve duplicate function detection to handle functions without function keyword

ichoosetoaccept:
The previous implementation assumed that function names would always be in the second field when using awk to extract them. This caused issues when the function keyword was omitted, as the function name could be in any field.

This fix:

1. Scans all fields in each line for a pattern matching a test function name
2. Uses a proper regex pattern to identify function names
3. Only processes and prints actual function names
4. Adds an explicit return 0 for clarity

Added test case to verify the fix works with function declarations that omit the function keyword.


Chemaclass (approved):
Awesome!

PR #399, opened 2025-05-21
fix: avoid printf error on floating-point values by using awk

drupol:
## 📚 Description

Bash's built-in `printf` does not support floating-point formats like `%f`. This causes errors such as 'invalid number' when formatting values like `167.183`.

## 🔖 Changes

- This commit update the usage of `printf` with `awk`, which correctly handles float rounding and improves portability across platforms including GNU Bash.

## ✅ To-do list

### DIFF
--- a/src/console_results.sh
+++ b/src/console_results.sh
@@ -106,7 +106,7 @@ function console_results::print_execution_time() {
     return
   fi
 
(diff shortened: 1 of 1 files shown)


Chemaclass (approved):

drupol:
Should I do something else here?

Chemaclass:
@drupol nope, all good, thanks

PR #416, opened 2025-06-02
Cache downloads endpoint 24h in localStorage

Chemaclass:
## 📚 Description

The homepage is doing a API call to get all download graph

Screenshot 2025-06-02 at 17 22 02

## 🔖 Changes

- Cache downloads endpoint 24h in localStorage

## ✅ To-do list

## 📹 Demo

https://github.com/user-attachments/assets/f25715f2-f68e-4b61-9a25-f7af2d9996a3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix typo "to has been called"
 - Add weekly downloads to the docs
+- Cache weekly downloads on localStorage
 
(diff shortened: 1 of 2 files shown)


Chemaclass:
After talking to @antonio-gg-dev we agreed is better to add cache-control header to the installer BE logic. No need to have this logic on the FE.

Issue #247, opened 2024-02-29
use tput or similar


antonio-gg-dev:
TODO #245

h0adp0re:
As the pioneer of this topic, I can add that after implementing `tput` and patching some codes into `assert_equals_ignore_colors`, really only the snapshots failed across different terminal emulators, since they capture everything, including the environment-specific stuff.

---

In addition to already mentioned benefits, a specific scenario improved via the use of `tput` that is probably not too specific to me — committing from inside of neovim (via `fugitive.vim`, for example).
This is the pre-commit hook output in the message area with the hardcoded colors:

```plain
lib/bashunit
^[[1m^[[32mbashunit^[[0m - 0.10.1
Running scripts/__tests__/example_test.sh
^[[33m↷ Skipped^[[0m: Skipped test example
^[[2m    Not needed in this env^[[0m
^[[36m✒ Incomplete^[[0m: Incomplete test example
^[[2m    Add unit tests for scripts^[[0m
^[[32m✓ Passed^[[0m: Passing test example

^[[2mTests:     ^[[0m ^[[32m1 passed^[[0m, ^[[33m1 skipped^[[0m, ^[[36m1 incomplete^[[0m, 3 total
^[[2mAssertions:^[[0m ^[[32m1 passed^[[0m, ^[[33m1 skipped^[[0m, ^[[36m1 incomplete^[[0m, 3 total
^[[46mSome tests incomplete^[[0m
```

This is the output with `tput` implemented:

```plain
lib/bashunit
bashunit - 0.10.1
Running scripts/__tests__/example_test.sh
↷ Skipped: Skipped test example
    Not needed in this env
✓ Passed: Passing test example
✒ Incomplete: Incomplete test example
    Add unit tests for scripts

Tests:      1 passed, 1 skipped, 1 incomplete, 3 total
Assertions: 1 passed, 1 skipped, 1 incomplete, 3 total

 Some tests incomplete
```

This behavior is largely explained in the first post under "Portability" but to add more resolution — in a limited environment where terminal output is not expected or needed, `$TERM` is (hopefully) set to `dumb`, which means no fancy output please. This ensures exactly what my example highlights: _just in case_ there happens to be output, make it still readable by not printing any color codes.

h0adp0re:
> I also have an observation to make after trying out tput, the GitHub actions went from being colored to black and white.

Yes, that is the effect of `TERM` being unset in that environment, which `tput` relies on to output _anything_.

[...]

PR #430, opened 2025-06-09
No colors when term is dumb

Chemaclass:
## 📚 Description

Closes: https://github.com/TypedDevs/bashunit/issues/247


Chemaclass:
Still dont know how to make this work for now lol

Issue #447, opened 2025-07-21
Failing tests on 0.22.0


drupol:
Hello,

| Q                | A                        |
|------------------|--------------------------|
| OS               | NixOS |
| Shell & version  | bash 3.2 |
| bashunit version | 0.22.0                    |

#### Summary

Today I tried to provide an update of Bashunit in Nix and while making the update (follow the PR at https://github.com/NixOS/nixpkgs/pull/427198), I noticed that 2 core tests are broken.

```
      >
       > Running tests/unit/test_doubles_test.sh
       > ✓ Passed: Successful mock                                                                18.818 ms
       > ✓ Passed: Successful override ps with echo with mock                                    16.3423 ms
       > ✓ Passed: Successful spy                                                                35.4721 ms
       > ✓ Passed: Unsuccessful spy called                                                       31.5284 ms
       > ✓ Passed: Successful spy called times                                                   30.3163 ms
       > ✓ Passed: Unsuccessful spy called times                                                 36.0456 ms
       > ✓ Passed: Successful spy with source function                                           38.8206 ms
       > ✓ Passed: Unsuccessful spy with source function have been called                        50.1873 ms
       > ✓ Passed: Successful spy called times with source                                        34.303 ms
       > ✓ Passed: Spy called in subshell                                                         55.583 ms
       > ✓ Passed: Mock called in subshell                                                        33.441 ms
       > ✓ Passed: Spy called with different arguments                                           80.8192 ms
       > ✓ Passed: Spy successful not called                                                     65.9594 ms
       > ✓ Passed: Spy unsuccessful not called                                                   77.5631 ms
       >
       > There were 2 failures:
       >
       > |1) tests/unit/assert_snapshot_test.sh
       > |✗ Failed: Assert match snapshot with placeholder
       > |    Expected 'to be empty'
       > |    but got  '✗ Failed: Assert match snapshot with placeholder
       > |    Expected to match the snapshot'
       > |2) tests/unit/assert_snapshot_test.sh
       > |✗ Failed: Assert match snapshot with custom placeholder
       > |    Expected 'to be empty'
       > |    but got  '✗ Failed: Assert match snapshot with custom placeholder
       > |    Expected to match the snapshot'
       >
       > Tests:      366 passed, 11 skipped, 7 incomplete, 2 snapshot, 2 failed, 388 total
       > Assertions: 510 passed, 11 skipped, 8 incomplete, 2 snapshot, 2 failed, 533 total
       >
[...]

PR #448, opened 2025-07-21
fix: prevent writing in source directory during tests

drupol:
This fix https://github.com/TypedDevs/bashunit/issues/447


Chemaclass (approved):
Woho! I will create a new realease tomorrow with this patch. Many thanks

PR #452, opened 2025-07-26
Fix broken core snapshot tests

Chemaclass:
## 🔖 Changes

- Updated snapshot path handling so that both snapshot assertion functions work correctly when tests are run using absolute paths. The directory path is now derived without a prefixed “./”

## ✅ To-do list


drupol:
However, the new CI workflow is useless, sadly... it's a bit of a waste of resources too to keep it there.

PR #453, opened 2025-07-26
Improve placeholder snapshot fallback without perl

Chemaclass:
## 🔖 Changes

- Updated the snapshot placeholder logic so custom placeholders now work without Perl by sanitizing the snapshot and converting the token to “.*” before constructing the fallback regex

- Added a test that mocks the absence of Perl to verify that custom placeholders still match correctly in this scenario

## ✅ To-do list


Chemaclass:
not really working out 🤔

PR #470, opened 2025-08-02
Lower min support bash 3.0

Chemaclass:
## 📚 Description

Right now the min supported version is bash 3.2

> Idea from @akinomyoga, thanks!

## 🔖 Changes

- Allow bash 3.0 and above

## ✅ To-do list


src/helpers.sh
```diff
@@ -239,7 +239,7 @@ function helpers::find_total_tests() {
                 for fn_name in "${functions_to_run[@]}"; do
                     local provider_data=()
```
akinomyoga, line 240: In Bash 3.0, this doesn't work. This is equivalent to `local provider_data='()'`.

akinomyoga (commented):
Have you tested this with a real Bash 3.0 binary?

src/helpers.sh
Chemaclass: Thanks for the feedback. I didn't tested with real bash 3.0 binary as I had to leave after I created the PR and wanted to know how far away from the final result this would be. I will continue with it and test it with bash 3.0 binary this weekend. Also, I should find a way to setup this bash 3.0 in the CI to run all tests using that version

Issue #478, opened 2025-09-02
Data providers do not work with empty strings or values containing whitespace


carlfriedrich:
| Q                | A                        |
|------------------|--------------------------|
| OS               | Linux on WSL |
| Shell & version  | bash 5.2.37 |
| bashunit version | 0.23.0                    |

#### Summary

First of all, thanks a lot for implementing and maintaining bashunit! I have been using it for a few days now and really like it so far.

There is one issue I stumbled upon today, though: data providers currently support neither empty strings nor values containing whitespaces.

The examples below are artificial, but I have a real-world use-case where I need to test a function with an empty string as an input, which is not possible unfortunately.

#### Current behavior

Sorrouding quotes around values are ignored. See two examples:

- Values containing spaces
  ```bash
  function data_provider_values() {
    echo "an input with spaces" "input2"
  }
  ```
  
  Expected values in the test function:
  - `$1`: `an input with spaces`
  - `$2`: `input2`

  Actual values in the test function:
  - `$1`: `an`
  - `$2`: `input`

- Empty values
  ```bash
  function data_provider_values() {
    echo "" "input"
  }
  ```

  Expected values in the test function:
  - `$1`: ``
  - `$2`: `input`

  Actual values in the test function:
  - `$1`: `input`

#### How to reproduce

```bash
# @data_provider data_provider_values
function test_data_provider() {
  local input="$1"
  local expected="$2"

  assert_equals "result" "${expected}"
}

function data_provider_values() {
  echo "input" "result"
  echo "some more input" "result"
  echo "" "result"
}
```

<img width="850" height="176" alt="Image" src="https://github.com/user-attachments/assets/b4ed2f0e-6c90-4be8-a5b6-33abc3cf2891" />

#### Expected behavior

Data providers should respect quotes to support values containing whitespaces or empty values.

carlfriedrich:
Fixed with #479.

PR #479, opened 2025-09-02
Fix: support whitespace in data providers

carlfriedrich:
## 📚 Description

This PR is a potential fix for #478. It adds a solution for supporting whitespace in data provider values.

## 🔖 Changes

The PR adds a global `data_set` function for declaring data sets in data providers, superseding the previous `echo` solution with a more robust way preserving quotes.

It furthermore changes the way data from data providers is parsed.
Previously the values were simply read using `IFS=" "`, which split the
input on spaces. This made values containing spaces impossible.
Now values are parsed using a combination of `eval` and `set`, which
respects quotes in the input data and hence makes contained spaces
possible. We can also provide empty strings as input data now.
The new parsing is backwards-compatible, so declaring data values using `echo` still works like before.

## ✅ To-do list

### DIFF
--- a/src/console_results.sh
+++ b/src/console_results.sh
@@ -132,7 +132,15 @@ function console_results::print_successful_test() {
   if [[ -z "$*" ]]; then
     line=$(printf "%s✓ Passed%s: %s" "$_COLOR_PASSED" "$_COLOR_DEFAULT" "$test_name")
   else
-    line=$(printf "%s✓ Passed%s: %s (%s)" "$_COLOR_PASSED" "$_COLOR_DEFAULT" "$test_name" "$*")
+    local quoted_args=""
+    for arg in "$@"; do
+      if [[ -z "$quoted_args" ]]; then
+        quoted_args="'$arg'"
+      else
+        quoted_args="$quoted_args, '$arg'"
+      fi
+    done
+    line=$(printf "%s✓ Passed%s: %s (%s)" "$_COLOR_PASSED" "$_COLOR_DEFAULT" "$test_name" "$quoted_args")
   fi
 
   local full_line=$line
--- a/src/globals.sh
+++ b/src/globals.sh
@@ -102,3 +102,18 @@ function print_line() {
   local char="${2:--}"      # Default to '-' if not passed
   printf '%*s\n' "$length" '' | tr ' ' "$char"
 }
+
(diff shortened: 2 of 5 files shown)


carlfriedrich:
Hey maintainers. 👋 I am willing to update changelog and documentation, but first I would like to get a first review on my approach for this fix.

Specific questions I have:

1. Do you like the new global function `data_set` for declaring data values in data provider functions? Is the naming appropriate?
2. Is the new parsing mechanism using `eval` suitable or do you have any concerns about it?
   Generally speaking, `eval` bears the risk of command injection. However, when used in conjunction with `data_set`, all input tokens are quoted, so this should be safe IMO, as long as we update the documentation to consequently use `data_set` instead of `echo` in data providers.

I am happy to hear your feedback on this. If we can agree on this solution I will add commits to update changelog and documentation.

Chemaclass (approved):
Awesome work, thank you!

Chemaclass:
@carlfriedrich please **update the docs** and add one liner to the `CHANGELOG.md` before merging this PR

carlfriedrich:
@Chemaclass 

I added documentation and changelog. I also added a commit which updates all bashunit test cases to use the new `data_set` function in data providers. This led to one other failing test case, which I pushed a fix for. I am not sure why this particular test is implemented the way it is (why do we [use a subshell to find the total tests](https://github.com/TypedDevs/bashunit/blob/05d5894946db498fd60268a82ea0890a59693fbe/tests/unit/helpers_test.sh#L260)?), that's why figuring out what was going on here took quite some time.

However, from my side this is merge-ready now.

Chemaclass (approved):
nice docs!

carlfriedrich:
@Chemaclass Thanks a lot for your fast and valuable feedback! Was fun contributing to this project. Looking forward to using it even more. 😃 
Can you already estimate when this PR will land in a release?

Chemaclass:
I think I will create a new release with the changes we already have by this weekend or next weekend. Until then, you can already benefit from the latest changes installing the beta (which is installing the code from main branch). 
> Here is how: https://bashunit.typeddevs.com/installation#define-custom-tag-and-folder

carlfriedrich:
Ah, great to know! Thanks a lot again!

PR #484, opened 2025-09-06
Fix: clean parallel temp folder after test run

carlfriedrich:
## 📚 Description

This PR resolves an issue where temporary files and directories bashunit creates implicitly during a parallel run were not properly cleaned up.

It adds both a unit test an an acceptance test that fail without the fix and now pass.

This issue already existed before my changes in #483 .

## 🔖 Changes

- Add a new function `parallel::cleanup` and call it in each exit point when parallel mode is enabled

*Side note*: the previous `parallel::reset` function (which I renamed to `parallel::init` now) contained the following line:

```bash
[ -f "$TEMP_FILE_PARALLEL_STOP_ON_FAILURE" ] && rm "$TEMP_FILE_PARALLEL_STOP_ON_FAILURE"
```

IMO the condition `[ -f "$TEMP_FILE_PARALLEL_STOP_ON_FAILURE" ]` could never be true, as the containing directory is deleted right before the line. Hence I removed it. If it served any purpose I overlooked, please tell me.

## ✅ To-do list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add support for empty values and values containing spaces in data providers via new `data_set` function
 - Document workaround for global function name collisions when sourcing scripts in tests by copying the original function
 - Fix `temp_dir` and `temp_file` data not being cleaned up when created in `set_up_before_script`
+- Fix `/tmp/bashunit/parallel` not being cleaned after test run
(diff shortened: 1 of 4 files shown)


Chemaclass (approved):
Nice catch. Thank you!

PR #487, opened 2025-09-07
Fix: support newlines in data providers

### LINKED ISSUE #486: Data providers do not work with strings containing newlines
| OS               | Linux on WSL  |
| Shell & version  | bash 5.2.37 |
| bashunit version | current main                 |

#### Summary

In #479 we added the new `data_set` function to support whitespace values in data providers. The function was implemented in a backwards-compatible way, still supporting the previous `echo`-based syntax.

The `data_set` function does not support data containing newlines, though. This test fails:

```bash
# @data_provider provide_value_with_newline
function test_value_with_newline_from_data_provider() {
  local value="$1"

  assert_same "value
with
newline" "$value"
}

function provide_value_with_newline() {
  data_set "value
with
newline"
}
```

This is the error:

```shell
✗ Failed: Value with newline from data provider
    Expected 'value
with
newline'
    but got  'value'
```

This is due to the backwards compatibility: the new parser still outputs the values line by line, just like `echo` did. This makes values with newlines impossible, unfortunately, because each new line is seen as a new input for the test.

I'd like to start a discussion if we could drop support for the previous `echo` syntax in order to support newline characters in `data_set` as well. I will definitely need this in the future, but I don't want to break anyone's workflow, and this definitely would be a breaking change.

Does anybody have an idea how to handle this? Are breaking changes okay while we're in major version 0? Or could we design this in a non-breaking way?

carlfriedrich:
## 📚 Description

This PR adds support for newlines in `data_set` values in data providers.

It also adds a unit test that fails without the fix and now passes.

Fixes #486.

## 🔖 Changes

- Use base64 encoding for passing data between the data provider parser and the test runner to preserve newlines and other special characters.

## ✅ To-do list


carlfriedrich:
I just realized that the previous limitation of a `data_set` for its last argument not to be empty is not present anymore with this fix, so I removed the corresponding note from the docs.

carlfriedrich:
This is ready for review now.

src/helpers.sh
```diff
@@ -61,6 +61,16 @@ function helper::interpolate_function_name() {
...
 
+function helper::encode_base64() {
+  local value="$1"
+
+  if command -v base64 >/dev/null; then
+    echo "$value" | base64 | tr -d '\n'
+  else
+    echo "$value" | openssl enc -base64 -A
+  fi
+}
```
Chemaclass, line 72: Nice one!

Issue #468, opened 2025-07-27
Add support for Bash 3.0


Chemaclass:
Right now, bashunit requires/support **Bash 3.2** and newer. However, why not expand the compatibility to start supporting from Bash 3.0 and so give that old code the possibility to write tests as well?

<img width="1071" height="409" alt="Image" src="https://github.com/user-attachments/assets/16460a24-9c32-4dc8-a8f3-6b354c2957ae" />

The reason why is currently 3.2 is because my own MacOS got this version by default. This has prevented me to use modern Bash syntax, therefore adapting to 3.0 might not be that difficult and we could give support to more developers out of the box.

PR #493, opened 2025-09-27
Lower bash support from 3.2 to 3.0

Chemaclass:
## 📚 Description

Closes: https://github.com/TypedDevs/bashunit/issues/468 by @akinomyoga

Right now, the min supported version is bash 3.2

## 🔖 Changes

- Allow bash 3.0 and above

## ✅ To-do list

### DIFF
--- a/src/helpers.sh
+++ b/src/helpers.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2004
+
 declare -r BASHUNIT_GIT_REPO="https://github.com/TypedDevs/bashunit"
 
 #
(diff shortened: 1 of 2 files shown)


src/assert.sh
```diff
@@ -546,7 +548,7 @@ function assert_line_count() {
     local actual
     actual=$(echo "$input_str" | wc -l | tr -d '[:blank:]')
     local additional_new_lines
-    additional_new_lines=$(grep -o '\\n' <<< "$input_str" | wc -l | tr -d '[:blank:]')
+    additional_new_lines=$(echo "$input_str" | grep -o '\\n' | wc -l | tr -d '[:blank:]')
```
akinomyoga: What does this change try to fix? `<<< "$input_str"` works in Bash 3.0.

PR #498, opened 2025-10-03
fix(data_set): preservation of trailing whitespace

### LINKED ISSUE #497: data_set does not preserve trailing whitespace in final argument
| OS               | macOS and Linux  |
| Shell & version  | bash 5.3.3 (macOS) and bash 5.2.32 (Linux) |
| bashunit version | 0.24.0 (both)                  |

#### Summary

`data_set` does not preserve trailing whitespace present in the final argument passed to it.

#### How to reproduce

This test currently fails:
```bash
# @data_provider provide_value_with_trailing_whitespace
function test_trailing_whitespace_in_last_value_from_data_provider() {
  local expected="$1"
  local actual="$2"

  assert_same "${expected}" "${actual}"
}

function provide_value_with_trailing_whitespace() {
  # Each data_set is passed the same value twice (expected, actual) to verify preservation
  data_set "value " "value "
  data_set "value    " "value    "
}
```

Output:
```
bashunit - 0.24.0 | Tests: 2
FF

There were 2 failures:

|1) tests/functional/provider_test.sh:107
|✗ Failed: Trailing whitespace in last value from data provider
|    Expected 'value '
|    but got  'value'
|2) tests/functional/provider_test.sh:107
|✗ Failed: Trailing whitespace in last value from data provider
|    Expected 'value    '
|    but got  'value   '



Tests:      2 failed, 2 total
Assertions: 2 failed, 2 total

 Some tests failed 
Time taken: 217 ms
```

#### Expected behavior

Test should pass

mattness:
## 📚 Description

Fixes #497

### Problem

`data_set` was not preserving trailing whitespace in its final argument. For example, `data_set "value "` or `data_set "value    "` would lose lose one space by the time the test function received the values

### Root Cause

The issue occurred in how `data_set` output was parsed. When using `eval "args=($output)"` to split the escaped arguments back into an array, bash's word splitting would trim trailing whitespace from the last element because there was nothing after it to delimit where it ended.

### Solution

Added an empty string sentinel (`''`) at the end of `data_set` output. This ensures there's always something after the last real argument, preventing bash from trimming its trailing whitespace during word splitting.

## 🔖 Changes

- Modified `data_set` to append an empty sentinel: `printf '%q\n' ""`
- Updated `runner::parse_data_provider_args` to remove the sentinel after parsing
- Added metacharacter detection to safely handle special characters (`*, |, &, ;`) that could break `eval`, falling back to character-by-character parsing when needed

## ✅ To-do list

### DIFF
--- a/src/runner.sh
+++ b/src/runner.sh
@@ -122,10 +122,11 @@ function runner::parse_data_provider_args() {
   fi
 
   # Try eval first (needed for $'...' from printf '%q'), unless metacharacters present
-  if [[ "$has_metachar" == false ]] && eval "args=($input)" 2>/dev/null; then
+  if [[ "$has_metachar" == false ]] && eval "args=($input)" 2>/dev/null && [[ ${#args[@]} -gt 0 ]]; then
     # Successfully parsed - remove sentinel if present
-    if [[ ${#args[@]} -gt 0 && -z "${args[-1]}" ]]; then
-      unset 'args[-1]'
+    local last_idx=$((${#args[@]} - 1))
+    if [[ -z "${args[$last_idx]}" ]]; then
+      unset 'args[$last_idx]'
     fi
     # Print args and return early
     for arg in "${args[@]}"; do


src/globals.sh
```diff
@@ -128,5 +128,5 @@ function data_set() {
       printf ' %q' "$arg"
     fi
   done
-  printf '\n'
+  printf ' %q\n' ""
 }
```
mattness, line 132: This _could_ be made more succinct with the following, but I chose not to do so in order to minimize what changed.
```
function data_set() {
  for arg in "${@}"; do
    printf '%q ' "${arg}"
  done
  printf '%q\n' ""  # Sentinel
}
```

Chemaclass (approved):
Thank you!

mattness:
Bash 3.x got me! Thanks for the fix :)

PR #513, opened 2025-11-22
Errors setup teardown still not caught correctly

### LINKED ISSUE #512: Errors in set_up and tear_down functions are still not caught correctly
| OS               | Linux |
| Shell & version  | bash 5.3.3 |
| bashunit version | 0.26.0                    |

#### Summary

This is a follow-up to #489. Parts of the issues were already addressed in #491. However, some errors are still not caught correctly.

#### Current behavior

While *non-existing* commands in `set_up` and `tear_down` functions are correctly logged, *failing* commands are still silently ignored.

#### How to reproduce

```bash
function set_up() {
  git checkout non-existing-ref
}

function test_asserting_foo_strings() {
  assert_same "foo" "foo"
}
```

Image

```bash
function set_up() {
  ls -la /non/existing/path
}

function test_asserting_foo_strings() {
  assert_same "foo" "foo"
}
```

Image

The only actual command resulting in an error is `false`, and only when it is the last statement in the function:

```bash
function set_up() {
  false
}

function test_asserting_foo_strings() {
  assert_same "foo" "foo"
}
```

Image

Or as I already wrote commands that do not exist:

```bash
function set_up() {
  non-existing-command
}

function test_asserting_foo_strings() {
  assert_same "foo" "foo"
}
```

Image

#### Expected behavior

`bashunit` should report failing commands in `set_up` and `tear_down` as well as `set_up_before_script` and `tear_down_after_script` functions.

Also IMO errors in `set_up` or `set_up_before_script` should prevent the test function from running at all.

Chemaclass:
## 📚 Description

Closes #512

Fixed lifecycle hooks not catching failing commands (non-zero exit codes) in `set_up`, `tear_down`, `set_up_before_script`, and `tear_down_after_script`.

**Problem:** Failing commands like `git checkout non-existing-ref`, `ls /non/existing/path`, or `false` were silently ignored in lifecycle hooks, allowing tests to run even when setup failed.

**Root cause:** Using `if ! command; then status=$?` - the `$?` inside the if block returns 0 (success of the conditional) instead of the command's actual exit code.

**Solution:** Changed to `command || exit $?` to capture the exit code immediately.

## 🔖 Changes

- Fixed `src/runner.sh:349` - Changed `if ! runner::run_set_up` to `runner::run_set_up || exit $?`
- Added test coverage for all lifecycle hooks with failing commands
- Updated CHANGELOG.md

### The Real Fix

  Both set_up and tear_down were broken for the same reason - they both call runner::execute_test_hook which was working correctly. The actual bug that affected BOTH was that runner::execute_test_hook wasn't catching failing commands.

  But once `runner::execute_test_hook` returns the correct error code:
  - `set_up` needed the fix at line 349 to properly exit on error
  - `tear_down` was already correctly using `|| teardown_status=$?` in the trap

  So both are fixed now, but through different mechanisms:
  1. The common fix: `runner::execute_test_hook` now correctly detects all failing commands
  2. The `set_up fix: || exit $?` ensures we exit when setup fails
  3. The `tear_down` "fix": It was already correctly written with `|| teardown_status=$?`

## ✅ To-do list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fix lifecycle hooks not catching failing commands (exit code errors)
+
 ## [0.26.0](https://github.com/TypedDevs/bashunit/compare/0.25.0...0.26.0) - 2025-11-02
 
 - Add `assert_unsuccessful_code` assertion to check for non-zero exit codes
--- a/src/runner.sh
+++ b/src/runner.sh
@@ -346,10 +346,7 @@ function runner::run_test() {
     # shellcheck disable=SC2064
     trap 'exit_code=$?; runner::cleanup_on_exit "$test_file" "$exit_code"' EXIT
     state::initialize_assertions_count
-    if ! runner::run_set_up "$test_file"; then
-      status=$?
-      exit "$status"
-    fi
+    runner::run_set_up "$test_file" || exit $?
 
     # 2>&1: Redirects the std-error (FD 2) to the std-output (FD 1).
     # points to the original std-output.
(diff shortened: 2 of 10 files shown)


Chemaclass:
@carlfriedrich, thanks for the feedback, I implemented the fix here https://github.com/TypedDevs/bashunit/pull/515 | [release 0.27](https://github.com/TypedDevs/bashunit/releases/tag/0.27.0)

PR #584, opened 2026-02-04
Support Bash 3.0

Chemaclass:
## Summary

Closes: https://github.com/TypedDevs/bashunit/issues/468 by @akinomyoga

Lower the minimum Bash version requirement from 3.2 to 3.0.

## Changes

- Replace Bash 3.1+ features with 3.0 compatible alternatives
- Fix `assert_contains_ignore_case` to use `tr` instead of `shopt nocasematch`
- Handle empty strings in `printf '%q'` and data providers
- Separate array declaration from assignment for Bash 3.0 compatibility
- Skip `release_test.sh` on Bash 3.0 (release.sh requires 3.1+)
- Add CI workflow to test all modes on Bash 3.0 in parallel

### DIFF
--- a/src/runner.sh
+++ b/src/runner.sh
@@ -769,7 +769,7 @@ function bashunit::runner::parse_result_parallel() {
   mkdir -p "$test_suite_dir"
 
   local sanitized_args
-  sanitized_args=$(echo "${args[*]}" | tr '[:upper:]' '[:lower:]' | sed -E 's/[^a-z0-9]+/-/g; s/^-|-$//')
+  sanitized_args=$(echo "${args[*]+"${args[*]}"}" | tr '[:upper:]' '[:lower:]' | sed -E 's/[^a-z0-9]+/-/g; s/^-|-$//')
   local template
   if [[ -z "$sanitized_args" ]]; then
     template="${fn_name}.XXXXXX"


src/runner.sh
```diff
@@ -305,25 +334,35 @@ function bashunit::runner::call_test_functions() {
       break
     fi
 
-    local provider_data=()
+    # Bash 3.0 compatible: unset before redeclaring to clear previous iteration's data
+    unset provider_data
```
akinomyoga: One should declare `provider_data` before the loop. Otherwise, this may remove user's global variable `provider_data` or a shell function `provider_data` if any.

src/state.sh
```diff
@@ -253,7 +253,7 @@ function bashunit::state::calculate_total_assertions() {
   numbers=$(echo "$input" | grep -oE '##ASSERTIONS_\w+=[0-9]+' | grep -oE '[0-9]+')
 
   for number in $numbers; do
```
akinomyoga: This isn't an issue with this PR, but the variable `number` leaks here.

Chemaclass: thanks! I added `local number`

akinomyoga: I quicly checked the codebase, but there seem to be many places where the loop variables are leaked.

akinomyoga (changes requested):
The regular expressions used in the following lines do not work in an intended way in Bash < 3.2.

```bash
src/coverage.sh:516:        if [[ $brace_count -eq 0 && "$line" =~ \{ && "$line" =~ \} ]]; then
install.sh:61:  [[ "$1" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ || "$1" == "latest" || "$1" == "beta" ]]
release.sh:532:  if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
src/learn.sh:861:  [[ $1 =~ ^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$ ]]
```

It seems you assign the regular expression to a variable and use it as `[[ ... =~ $variable ]]`, but I recommend you to define a function like `bashunit::regex_match() { [[ $1 =~ $2 ]]; }` so that you can call `bashunit::regex_match "$line" '\{'`, etc.

src/coverage.sh
```diff
@@ -363,13 +366,17 @@ function bashunit::coverage::is_executable_line() {
   [[ "$line" =~ ^[[:space:]]*[\{\}][[:space:]]*$ ]] && return 1
 
   # Skip control flow keywords (then, else, fi, do, done, esac, in, ;;, ;&, ;;&)
-  [[ "$line" =~ ^[[:space:]]*(then|else|fi|do|done|esac|in|;;|;;&|;&)[[:space:]]*(#.*)?$ ]] && return 1
+  # Pattern stored in variable for Bash 3.0 compatibility
+  local _ctrl_pattern='^[[:space:]]*(then|else|fi|do|done|esac|in|;;|;;&|;&)[[:space:]]*(#.*)?$'
+  [[ "$line" =~ $_ctrl_pattern ]] && return 1
```
akinomyoga: In the previous comment, I implied using `bashunit::regex_match` everywhere for this kind of manipulation, but is there a reason that the regular expression wants to be stored in a variable in this context? This applies to all the other occurrences of the pattern `local var='...'; ... ; [[ ... =~ $var ]]`. I think it should be normalized in either way; use variables everywhere or use `bashunit::regex_match` everywhere.

src/coverage.sh
```diff
@@ -429,7 +437,7 @@ function bashunit::coverage::get_line_hits() {
     return
   fi
 
-  local count
+  local count=0
```
akinomyoga: This is actually unnecessary.

src/coverage.sh
```diff
@@ -806,13 +824,17 @@ function bashunit::coverage::generate_index_html() {
   local tests_passed="$6"
   local tests_failed="$7"
   shift 7
-  local file_data=()
-  [[ $# -gt 0 ]] && file_data=("$@")
+  # Handle array passed as arguments - Bash 3.0 compatible
+  local IFS=$' \t\n'
```
akinomyoga: What is the purpose of this `IFS`? This line is below the comment *"# Handle array passed as arguments - Bash 3.0 compatible"*, but the immediately following codes are unaffected by `IFS`.

Or maybe this is to set the normal `IFS` for the entire processing in the function? In that case, I think this line should be put earlier in the function.

src/doc.sh
```diff
@@ -11,13 +11,16 @@ function bashunit::doc::get_embedded_docs() {
...
   local filter="${1:-}"
-  local line
+  local line=""
   local docstring=""
   local fn=""
   local should_print=0
 
+  # Pattern stored in variable for Bash 3.0 compatibility
+  local _doc_pattern='^## ([A-Za-z0-9_]+)'
+  local line
```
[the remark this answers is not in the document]
akinomyoga:
```suggestion
  local docstring=""
  local fn=""
  local should_print=0

  # Pattern stored in variable for Bash 3.0 compatibility
  local _doc_pattern='^## ([A-Za-z0-9_]+)'
  local line
```

Removing the former.

src/main.sh
```diff
@@ -185,8 +194,9 @@ function bashunit::main::cmd_test() {
           inline_filter_file="$parsed_path"
         fi
 
+        local file=""
```
akinomyoga:
```suggestion
        local file
```

src/main.sh
```diff
@@ -166,13 +173,15 @@ function bashunit::main::cmd_test() {
...
-  if [[ ${#raw_args[@]} -gt 0 ]]; then
+  if [[ "$raw_args_count" -gt 0 ]]; then
     if [[ -n "$assert_fn" ]]; then
       # Assert mode: pass args as-is without file path processing
       args=("${raw_args[@]}")
+      args_count="$raw_args_count"
     else
       # Test mode: process file paths and extract inline filters
-      for arg in "${raw_args[@]}"; do
+      local arg=""
```
akinomyoga:
```suggestion
      local arg
```

src/reports.sh
```diff
@@ -62,12 +62,13 @@ function bashunit::reports::generate_junit_xml() {
     echo "             skipped=\"$tests_skipped\" snapshot=\"$tests_snapshot\""
     echo "             time=\"$time\">"
 
+    local i=0
```
akinomyoga:
```suggestion
    local i
```

src/reports.sh
```diff
@@ -97,11 +98,12 @@ function bashunit::reports::generate_report_html() {
 
   # Collect test cases by file
   : > "$temp_file"  # Clear temp file if it exists
+  local i=0
```
akinomyoga:
```suggestion
  local i
```

src/assert.sh
```diff
@@ -665,18 +690,19 @@ function assert_line_count() {
   bashunit::assert::should_skip && return 0
 
   local expected="$1"
-  local input_arr=("${@:2}")
+  local -a input_arr=()
```
akinomyoga:
```suggestion
  local -a input_arr
```

src/runner.sh
```diff
@@ -106,9 +115,11 @@ function bashunit::runner::load_test_files() {
 function bashunit::runner::load_bench_files() {
   local filter=$1
   shift
-  local files=("${@}")
+  local -a files=()
+  [[ $# -gt 0 ]] && files=("$@")
 
-  for bench_file in "${files[@]}"; do
+  local bench_file
+  for bench_file in ${files+"${files[@]}"}; do
```
akinomyoga: `IFS` is not set in this function too.

src/assert.sh
```diff
@@ -665,18 +690,19 @@ function assert_line_count() {
...
+  input_arr=("${@:2}")
   local input_str
-  input_str=$(printf '%s\n' "${input_arr[@]}")
+  input_str=$(printf '%s\n' ${input_arr+"${input_arr[@]}"})
```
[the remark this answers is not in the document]
akinomyoga: I cannot close this conversation (sub-thread) because I don't have the right in this repository. Could you close the conversation?

src/runner.sh
akinomyoga: This can also be closed now.

PR #599, opened 2026-03-06
feat: P1 developer experience improvements

Chemaclass:
## Summary

Developer experience improvements focused on workflow and debugging:

- **`--jobs N` flag** — Limit parallel test concurrency to N jobs (e.g., `--jobs 4`)
- **`--watch` mode** — Automatically re-run tests when files change (`--watch`)
- **Source context in failures** — Show the assertion lines from test source when a test fails
- **TAP output format** — TAP version 13 compliant output via `--output tap` for CI/CD integration

## Commits

1. `feat(cli): add --jobs N flag to limit parallel concurrency`
2. `feat(cli): add --watch mode for automatic test re-runs`
3. `feat(runner): show source context in failure summary`
4. `feat(cli): add TAP output format via --output tap`

## Test plan

- [x] All 819 tests passing (807 passed, 3 skipped, 7 incomplete, 2 snapshot)
- [x] ShellCheck passes on all modified files
- [x] EditorConfig lint passes
- [x] Pre-commit hook passes
- [x] TAP output verified with test files
- [x] Bash 3.0+ compatible (no associative arrays, `[[]]` used only where existing)

### DIFF
--- a/docs/command-line.md
+++ b/docs/command-line.md
@@ -56,6 +56,7 @@ bashunit test tests/ --parallel --simple
 | `-e, --env, --boot <file>`     | Load custom env/bootstrap file (supports args)   |
 | `-f, --filter <name>`          | Only run tests matching name                     |
 | `--log-junit <file>`           | Write JUnit XML report                           |
+| `-j, --jobs <N>`               | Run tests in parallel with max N concurrent jobs |
 | `-p, --parallel`               | Run tests in parallel                            |
 | `--no-parallel`                | Run tests sequentially                           |
 | `-r, --report-html <file>`     | Write HTML report                                |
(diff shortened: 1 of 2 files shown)


src/console_header.sh
```diff
@@ -109,11 +109,13 @@ Options:
   --tag <name>                Only run tests with matching @tag (repeatable, OR logic)
   --exclude-tag <name>        Skip tests with matching @tag (repeatable, exclude wins)
   --log-junit <file>          Write JUnit XML report
-  -p, --parallel              Run tests in parallel
+  -j, --jobs <N>              Run tests in parallel with max N concurrent jobs
```
JesusValeraDev: Is the documentation updated?

JesusValeraDev (approved):

src/console_header.sh
Chemaclass: Updated! Added `--jobs` documentation to both `docs/command-line.md` (options table + dedicated section) and `docs/configuration.md` (`BASHUNIT_PARALLEL_JOBS` env var).

PR #619, opened 2026-04-09
Fix post-increment causing silent exit under set -e

objctp:
## Background

Related #618 

LCOV and HTML coverage reports were incomplete/missing due to `((var++))` post-increment operator causing early script termination under `set -e`.

## Changes

- Fixed `report_lcov()` and `generate_file_html()` post-increment causing silent exit under `set -e`
- Standardised remaining `((var++))` to `((++var))` for consistency (preventive)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix spying on `echo` or `printf` causing bashunit to hang due to infinite recursion (#607)
 - Fix invalid `.env.example` coverage threshold entry and copy `.env.example` to `.env` in CI test workflows so configuration parse errors are caught during automated test runs
 - Fix `clock::now` shell-time parsing when `EPOCHREALTIME` uses a comma decimal separator
+- Fix LCOV and HTML coverage reports generating incomplete/empty output due to post-increment operator causing silent exit under `set -e` (#618)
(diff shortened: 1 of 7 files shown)


Chemaclass (commented):
- **Complete coverage**: all standalone `((var++))` outside `for` loop headers are changed. `state.sh` is correctly left alone since it already uses `|| true`.
- **Correct scope**: bashunit runs with `set -euo pipefail` (set in `/bashunit` line 2 and `src/globals.sh`), so this affects bashunit's own execution, not just user scripts.
- **CHANGELOG entry** is accurate and references #618.
- **No behavior change**: only the expression return value changes, not the increment itself.

### Concern: No regression test

The PR has zero tests. Given this bug causes **silent** failures (the `{ ... } > "$output_file"` redirect swallows the error), a regression test would be valuable. Something like:

```bash
function test_get_executable_lines_succeeds_under_set_e() {
  local result
  result=$(set -e; bashunit::coverage::get_executable_lines "some_file.sh")
  assert_greater_than 0 "$result"
}
```

This would prevent future regressions if someone reintroduces `((var++))`.

Chemaclass (approved):

Issue #284, opened 2024-07-12
Compare testsuite with baseline


Chemaclass:
Using the current logging system, we can generate a JUnit XML log file that could be as a baseline. Then, we could run a new flag to compare our current result to that baseline and output the exact different (if any) to be able to easily identify the difference test-wise between before and after wrt your testsuite.

Chemaclass:
After implementing, decided baseline is poor fit for shell-scale test suites. Permanent CLI surface not justified by adoption value. Use --filter / --exclude-tag for selective runs instead.

PR #645, opened 2026-04-29
feat(cli): generate baseline of known failures

Chemaclass:
## 🤔 Background

Related #284

Inspired by PHPUnit's baseline feature: capture currently failing/risky/incomplete tests so the suite can focus on newly introduced issues without rewriting or skipping legacy ones.

## 💡 Changes

- Add `--generate-baseline <file>` (and `BASHUNIT_BASELINE_GENERATE`) writing an XML list of failed/risky/incomplete tests; the run exits 0 so the file can be committed.
- Add `--use-baseline <file>` (and `BASHUNIT_BASELINE_USE`) which loads the XML and reports matching tests as `baselined`, excluded from the failure count; unmatched failures still fail the run.
- Console summary, help text, CHANGELOG and docs (command-line, configuration) updated; Bash 3.0+ compatible (parallel indexed arrays, no `[[ ]]`).


Chemaclass:
Reconsidered scope. Closing — see issue #284.

PR #646, opened 2026-04-29
feat(io): tput capability probe and centralized color escapes

### LINKED ISSUE #247: use tput or similar

Chemaclass:
## Summary

Closes #247.

Adopts a hybrid tput + ANSI strategy (see [ADR-006](adrs/adr-006-tput-vs-ansi-escapes.md)): keep ANSI as the primary emission mechanism via the existing `bashunit::sgr` helper, and use `tput` only for terminal capability probing and screen clearing — the two places where it actually buys portability without the per-emission failure modes that destabilized prior attempts (PR #245).

- `bashunit::env::supports_color` — returns false on `TERM=dumb` or when `tput colors` reports fewer than 8. Wired into `src/colors.sh` so colors auto-disable when the terminal cannot render them, complementing `--no-color` / `NO_COLOR`.
- `bashunit::io::clear_screen` — prefers `tput clear`, falls back to `\033[2J\033[H`. Replaces the hardcoded escape in `--watch` mode (`src/main.sh`).
- `src/coverage.sh` — hardcoded `\033[...m` escapes replaced with the central `_BASHUNIT_COLOR_*` constants. Single source of truth for color emission.
- ADR-006 documents the decision and rejected alternatives.

No public API changes. Behavior matches status quo on color-capable terminals; on dumb terminals, color now auto-disables instead of emitting raw escapes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,11 @@
 
 ### Added
 - Display captured test output on assertion failures when `--show-output` is enabled (#637)
+- Auto-disable colored output when the terminal cannot render it: `TERM=dumb` or `tput colors` reporting fewer than 8 now suppress color the same way `--no-color` / `NO_COLOR` does (#247)
+- `bashunit::env::supports_color` helper exposing the capability probe; `bashunit::io::clear_screen` helper that prefers `tput clear` and falls back to the raw ANSI sequence (#247)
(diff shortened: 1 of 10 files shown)


h0adp0re:
Thank you, this is excellent!

Issue #636, opened 2026-04-23
Speed up coverage report generation


RevanProdigalKnight:
As I continue to expand my own project's suite of unit tests I have started to look through the code in bashunit to try and identify pain points. I know I'm penalizing myself by running bashunit on a 15+ year old computer, in Cygwin, on Windows 10 (as opposed to running on proper Linux), as on my PC the test suite takes several minutes to run in serial mode and on the build pipeline agent, the same suite takes under a minute to run.

Even considering all of that, on my PC unit test generation times are... rather long. (I haven't tried introducing it on the build pipeline yet)

```sh
$ time ./lib/bashunit tests/* --simple --jobs 4 --coverage --coverage-report-html
bashunit - 0.34.1
# Test output omitted for brevity
Tests:      194 passed, 5 skipped, 2 incomplete, 201 total
Assertions: 1100 passed, 5 skipped, 2 incomplete, 1107 total

 Some tests incomplete
Time taken: 9m 14s

Coverage Report
---------------
# Specific file output omitted for brevity
---------------
Total: 1409/1734 (81%)


Coverage report written to: coverage/lcov.info
Coverage HTML report written to: coverage/html/index.html

real    217m36.766s
user    52m52.586s
sys     247m19.326s
```

Running with `--parallel` for unlimited testing concurrency instead of a fixed number of jobs does not increase speed significantly, since my PC is so old and only has 6 CPU cores.

As an aside, for some reason my unit tests ran faster *before* I split the main file into smaller modules, taking ~5 minutes in serial, but I was unable to get any coverage information out of it. Not sure how/why that is the case, but thought it was worth mentioning as an interesting tidbit.

---

In my testing so far, `bashunit::coverage::is_executable_line` can be optimized from running *up to* 7 subshells of `$(echo "$line" | "$GREP" [...] || true)` for the various omissions, to at most 1 with the following (this includes the changes made in `main` for handling loop terminators with redirections/etc. which haven't made it to a released version yet):

```sh
_BASHUNIT_NONEXECUTABLE_LINE_PATTERNS_ARRAY=(
  # Skip function declaration lines but not single-line functions with a body
  '(function[[:space:]]+)?[a-zA-Z_][a-zA-Z0-9_:]*[[:space:]]*\(\)[[:space:]]*\{?'
  # Skip lines with only braces
  '[\{\}]'
  # Skip control flow keywords
  '(then|else|fi|do|done|esac|in|;;|;;&|;&)'
[...]

PR #647, opened 2026-04-30
Further profiling and optimisation #636

objctp:
## Background

Related #636

PRs #643 and #644 addressed the `is_executable_line` consolidation and `compute_file_coverage` single-pass. This PR adds further optimisations on top: eliminating remaining `sed`/`grep` subshells, pre-loading source files into arrays, caching stats across reports, and using pure Bash 3.0 string operations in `extract_functions` (replacing `[[ =~ ]]`).

## Changes

- **Eliminate `sed -n` per-line spawns** — read source files once into indexed arrays in `get_hit_lines` and `generate_file_html`, replacing per-line sed calls with array lookups
- **Cache file stats across reports** — `precompute_file_stats` computes executable/hit/pct once via `compute_file_coverage`, shared by `report_text`, `report_html`, and `get_percentage`; O(1) lookup with string-based cache
- **Bash 3.0 compatible `extract_functions`** — pure string operations replace `[[ =~ ]]` + `BASH_REMATCH`; `printf '%s'` replaces `echo` in `is_executable_line`; `((++var))` convention from #619
- **Pre-load file arrays in hot paths** — `compute_file_coverage`, `report_lcov`, and `get_function_coverage` now read files once and iterate with `for` instead of `while IFS= read`
- **Pure Bash fast-path in `is_executable_line`** — `case` statements handle comments, braces, control keywords, and standalone `)` without spawning grep; falls through to grep for complex patterns only
- Added tests for `precompute_file_stats` and `get_cached_stats`

### Per-Component Breakdown

> **Note:** BEFORE = original `main` before any optimisation PRs (#643, #644, and this PR). AFTER = this branch, which includes all three PRs' changes. The per-component and hyperfine numbers reflect the cumulative improvement.

Test workload: 15 source files, ~1200 hit entries, full report generation (text + lcov + html)

| Component                          | BEFORE         | AFTER         | Improvement  |
| ---------------------------------- | -------------- | ------------- | -------- |
| `is_executable_line` (8000 calls)  | 231,382 ms     | 30,556 ms     | 7.6x     |
| `extract_functions` (15 calls)     | 9,396 ms       | 93 ms         | 101.0x   |
| `report_text`                      | 74,480 ms      | 6,726 ms      | 11.1x    |
| `report_lcov`                      | 124,797 ms     | 6,334 ms      | 19.7x    |
| `report_html`                      | 213,923 ms     | 32,495 ms     | 6.6x     |
| **Full pipeline (text+lcov+html)** | **406,970 ms** | **44,797 ms** | **9.1x** |

### Hyperfine Statistical Benchmark

Measured with warmup run + 3 measured runs each, on a macOS machine:

```
Benchmark 1: BEFORE (main)
  Time (mean ± σ):     118.053 s ±  0.930 s
  Range (min … max):   117.169 s … 119.022 s    3 runs

Benchmark 2: AFTER (optimised)
  Time (mean ± σ):      8.782 s ±  0.209 s
  Range (min … max):    8.618 s … 9.018 s    3 runs

Summary
  AFTER ran 13.44 ± 0.34 times faster than BEFORE
```

**13.4x faster** confirmed by hyperfine. The original code took ~2 minutes to generate all three reports; after the optimisations it takes ~9 seconds.

### Caveats

- Numbers from macOS with moderate fork overhead. Cygwin environments have orders-of-magnitude higher fork costs, so subshell elimination gains would be amplified further.
- Per-component numbers use `EPOCHREALTIME` timestamps in a benchmark harness — directional rather than statistically rigorous. The hyperfine full-pipeline comparison is the headline number.
- Benchmark workload (15 files, ~1200 entries) is synthetic. Real-world speedup depends on repository size and hit density.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Speed up coverage report generation by collapsing the per-line non-executable pattern checks in `bashunit::coverage::is_executable_line` into a single combined `grep` invocation (#636)
 - Speed up coverage report generation further by combining executable + hit counting into a single source-file pass (`bashunit::coverage::compute_file_coverage`) shared across text/lcov/html reporters, removing per-line `get_line_hits` scans of the coverage data file (#636)
 - Replace `echo | sed` / `echo | grep` subshells in `bashunit::coverage::extract_functions` with bash native regex matching and parameter expansion (#636)
+- Speed up coverage report generation by replacing per-line `sed` lookups with pre-loaded indexed arrays in `get_hit_lines` and `generate_file_html` (#636)
+- Speed up coverage report generation by caching pre-computed file stats across text/lcov/html reports (#636)
 
 ## [0.35.0](https://github.com/TypedDevs/bashunit/compare/0.34.1...0.35.0) - 2026-04-26
 
--- a/src/coverage.sh
+++ b/src/coverage.sh
@@ -190,6 +190,57 @@ function bashunit::coverage::get_file_stats() {
   echo "${executable}:${hit}:${pct}:${class}"
 }
 
+# Pre-computed file stats cache (avoids redundant per-file reads across reports)
+_BASHUNIT_COVERAGE_STATS_FILES=()
+_BASHUNIT_COVERAGE_STATS_EXEC=()
+_BASHUNIT_COVERAGE_STATS_HIT=()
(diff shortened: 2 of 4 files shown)


Chemaclass (approved):
LGTM

PR #675, opened 2026-05-12
ref(bash): codify dynamic-scope safety for outvar helpers

### LINKED ISSUE #674: ref(bash): codify dynamic-scope safety for helpers that take variable names
## Context

Bash `local` is **dynamically scoped**, not lexically scoped. A `local` declared inside a callee is visible to (and shadows) any caller variable of the same name during the call. This bit us in #662 / PR #672:

```bash
function bashunit::runner::format_subshell_output() {
  local _out=$1
  local subshell_output=$2     # LOCAL shadows caller's same-named var
  ...
  eval "$_out=\$line"          # assigns to LOCAL, not caller
}

# Caller — passes its own var name as the outvar
local subshell_output="raw"
bashunit::runner::format_subshell_output subshell_output "$subshell_output"
echo "$subshell_output"        # still "raw" — silent data loss
```

That bug surfaced as 12 parallel test failures and was only fixed by renaming all internal locals with a `__bu_` prefix. Worth a sweep to confirm no similar landmines exist elsewhere and to codify the rule so future helpers don't regress.

## Audit findings

I scanned `src/*.sh` for the three relevant patterns:

### 1. Eval-assignment outvar (write back via parameter name)

Pattern: `eval "$out_var_name=..."` where `out_var_name` is `$1`.

| Location | Status |
|----------|--------|
| `src/runner.sh` — `extract_encoded_field`, `extract_subshell_type`, `format_subshell_output`, `compute_total_assertions` | Fixed in #672 with `__bu_` prefix |

No other eval-assignment outvar helpers exist today. Good baseline.

### 2. Indirect-read by parameter-constructed name (`${!var}`)

Pattern: function takes a value, constructs a variable name from it, reads via `${!constructed_name}`. Vulnerable when a caller's local happens to match the constructed name.

| Location | Construct | Risk |
|----------|-----------|------|
| `src/test_doubles.sh:21,22` — `bashunit::unmock` | `${variable}_times_file`, `${variable}_params_file` | Theoretical |
| `src/test_doubles.sh:94,95` — `assert_have_been_called` | `${variable}_times_file` | Theoretical |
| `src/test_doubles.sh:124,126,128` — `assert_have_been_called_with` | `${variable}_params_file` | Theoretical |
| `src/test_doubles.sh:152,153` — `assert_have_been_called_times` | `${variable}_times_file` | Theoretical |
| `src/test_doubles.sh:181,182,193,194` — `assert_have_been_called_nth_with` | `${variable}_times_file`, `${variable}_params_file` | Theoretical |

These resolve names like `mycmd_times_file` (derived from sanitized command name). A test that locally declares a variable with that exact derived name and then calls the spy/assert helper would silently read the local instead of the exported global. Real-world likelihood is low but the failure mode is invisible (no error, just wrong assertion).

### 3. Intentional dynamic-scope mutation

| Location | Notes |
|----------|-------|
| `src/coverage.sh:818-829` — `_branch_push_if`, `_branch_close_if_arm`, etc. | **Documented** in the comment block at lines 818-821 as intentional (helpers mutate caller's `if_decision_line`, `if_arms`, `if_depth`, `if_arm_start` arrays). Fine as-is; the rationale is recorded. |
| `src/coverage.sh:784` — `_BASHUNIT_BRANCH_ARMS_OUT` global slot | Single-slot global output; safe (no name collision possible, only one output channel). |

### Must

1. **Codify the convention in `.claude/rules/bash-style.md`** under a new section "Outvar helpers (dynamic-scope safety)". Required wording:
   - "Helpers that take a caller-named variable as an outvar (first arg) MUST prefix all internal locals with `__bu_` to avoid shadowing the caller's variable through dynamic scoping."
   - Include the minimal failing example (the `subshell_output` case above).
   - Mention that `printf -v` is not Bash 3.0-safe and `declare -n` requires 4.3+, hence the `eval "$_out=\$__bu_val"` + prefix pattern is the portable contract.

### Should

2. **Defensive audit of `src/test_doubles.sh`**: rename internal locals that participate in constructed names so a colliding caller local cannot be confused with the exported global. Concretely, the `variable`, `file_var`, `times_file_var`, `params_file_var` locals on the indirect-read paths. No bug filed today; this is hardening.

3. **Add a regression test** that exercises the shadow scenario for at least one outvar helper from #672: caller declares a local with the same name as one of the helper's internal locals, then calls the helper; assert the helper still produces the right value. Cheap insurance against someone dropping the prefix later.

### Nice to have

4. **shellcheck-style lint guard**: a small CI script that greps `src/*.sh` for `eval "\$[a-zA-Z_]*=` and asserts every match lives in a function whose locals are `__bu_`-prefixed. Catches the regression mechanically.

## Acceptance

- [ ] Rule documented in `.claude/rules/bash-style.md`
- [ ] `test_doubles.sh` indirect-read sites audited and either renamed or explicitly noted as safe
- [ ] At least one regression test for the shadow scenario in `tests/unit/runner_test.sh`
- [ ] Bash 3.0+ compat preserved (no namrefs, no `declare -n`)

Chemaclass:
## Summary
- Hot-path helpers in `src/runner.sh` return via dedicated global slots (`_BASHUNIT_RUNNER_*_OUT`) instead of stdout. No `$(...)` fork, no eval, no shadowing risk — caller-named locals are simply never touched.
- Spy state in `src/test_doubles.sh` is namespaced under `_BASHUNIT_SPY_*`, so a test that uses a `local foo_times_file=...` cannot collide with the framework's `_BASHUNIT_SPY_foo_TIMES_FILE`.
- `.claude/rules/bash-style.md` documents three patterns in priority order:
  1. **Preferred:** dedicated global return slot.
  2. **For helpers that build dynamic names:** namespace the constructed name.
  3. **Last resort:** outvar-by-name with `__bu_` prefix + regression test (none in tree).
- Regression tests in `tests/unit/runner_test.sh` confirm helpers do not touch caller-named locals.

## Why
Bash `local` is dynamically scoped, so any helper that touches caller-named variables (via `eval "$name=..."`, `printf -v`, `${!name}` or `export "$name"=...`) can silently corrupt a caller that picks a natural name. The fix is to avoid that pattern entirely: return through a fixed framework-owned slot, or namespace the names we construct.

## Test plan
- [x] `./bashunit tests/unit/` green (839 passed, 1m20s)
- [x] `./bashunit --parallel tests/` green (1074 passed, 46.58s)
- [x] `make sa` clean
- [x] `make lint` clean
- [x] Bash 3.0+ compat preserved (no `declare -n`, no `printf -v`)

Closes #674

### DIFF
--- a/.claude/rules/bash-style.md
+++ b/.claude/rules/bash-style.md
@@ -49,6 +49,77 @@ Constants -> Globals -> Private functions -> Public functions
 
 Source deps relative to script: `"$(dirname "${BASH_SOURCE[0]}")/dep.sh"`
 
+## Outvar helpers (dynamic-scope safety)
+
+Bash `local` is **dynamically scoped**, not lexically scoped. A `local` inside a callee
+shadows the caller's same-named variable for the duration of the call. Any helper that
+writes back to a caller-named variable (the **outvar pattern**) must defend against this
+or it will silently corrupt callers that pick a "natural" name.
+
+### Rule
+
+Helpers that take a caller-named variable as an outvar (typically the first argument)
+**MUST prefix every internal local with `__bu_`** so no caller-passed name can collide.
+
+### Why
+
+Without the prefix, this fails silently:
+
+```bash
+without_prefix() {
+  local _out=$1
+  local subshell_output=$2     # LOCAL shadows caller's same-named var
+  local line="formatted"
+  eval "$_out=\$line"          # assigns to the LOCAL, not the caller
+}
+
+main() {
+  local subshell_output="raw"
+  without_prefix subshell_output "$subshell_output"
+  echo "$subshell_output"      # prints "raw" — the helper appeared to succeed
+}
+```
+
+The caller's variable is untouched, no error is raised, and tests downstream silently
+get stale data. This regression bit us in #662 (12 parallel test failures, fixed in
+PR #672) before the prefix was applied.
+
+### Pattern
+
+```bash
+# Writes <description> into the named outvar.
+# Arguments: $1 outvar name, $2 input
+function bashunit::pkg::do_thing() {
+  local __bu_out=$1
+  local __bu_in=$2
+  local __bu_val="${__bu_in%%]*}"
+  __bu_val="${__bu_val#[}"
+  eval "$__bu_out=\$__bu_val"
+}
+```
+
+- Use `eval "$__bu_out=\$__bu_val"` for the assignment. Quote the `$` of the value with
+  a backslash so only the outvar name is expanded at eval time.
+- Do **not** use `printf -v` (Bash 3.0 lacks it) or `declare -n` (Bash 4.3+).
+- A regression test should pass the helper's own local names as outvar to catch any
+  contributor who later drops the prefix.
+
+### Where the convention applies in this repo
+
+- All four hot-path helpers in `src/runner.sh`: `extract_encoded_field`,
+  `extract_subshell_type`, `format_subshell_output`, `compute_total_assertions`.
+- Indirect-read sites in `src/test_doubles.sh` (`${!file_var}`) follow the same rule
+  defensively — internal locals are `__bu_`-prefixed even though the read direction is
+  less likely to collide in practice.
+
+### Intentional dynamic-scope mutation is a separate pattern
+
+The coverage branch helpers in `src/coverage.sh` (`_branch_push_if` and friends) deliberately
+mutate caller locals (`if_decision_line`, `if_arms`, `if_depth`, `if_arm_start`). That is
+documented inline at `src/coverage.sh:818-821` and is **not** the outvar pattern — the
+helper has no `$1`-named outvar argument; the caller agrees to share state by convention.
+Don't introduce new instances of this pattern without an inline justification comment.
+
 ## ShellCheck
 
 All code must pass `make sa`. Use directives sparingly with reason:
--- a/.claude/rules/bash-style.md
+++ b/.claude/rules/bash-style.md
@@ -49,70 +49,85 @@ Constants -> Globals -> Private functions -> Public functions
 
 Source deps relative to script: `"$(dirname "${BASH_SOURCE[0]}")/dep.sh"`
 
-## Outvar helpers (dynamic-scope safety)
+## Returning values from a helper (dynamic-scope safety)
 
 Bash `local` is **dynamically scoped**, not lexically scoped. A `local` inside a callee
-shadows the caller's same-named variable for the duration of the call. Any helper that
-writes back to a caller-named variable (the **outvar pattern**) must defend against this
-or it will silently corrupt callers that pick a "natural" name.
+shadows the caller's same-named variable for the duration of the call. Same trap fires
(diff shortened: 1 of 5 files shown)


JesusValeraDev (approved):

Chemaclass:
Good catch on the audit gap. Extended in 60a31d1:

- `bashunit::mock` and `bashunit::spy` now also use the `__bu_` prefix on all internal locals. Both helpers take a caller-named `command` and write derived globals via `export "${variable}_times_file"=...`. If a caller has a local matching that constructed name, `export NAME=value` updates the local in-place (dynamic scoping) instead of creating a clean env binding — same hazard class as the indirect-read sites.
- Rule note in `.claude/rules/bash-style.md` widened to cover the full `test_doubles.sh` surface (mock/spy/unmock + the assert_have_been_called family), not just the read direction.

Final audit state across `src/*.sh`:
| File | Pattern | Status |
|------|---------|--------|
| `src/runner.sh` | eval outvar | `__bu_` prefix (PR #672) |
| `src/test_doubles.sh` | export + `${!var}` | `__bu_` prefix (this PR) |
| `src/coverage.sh` | intentional dynamic-scope | documented exception, kept |
| `src/helpers.sh` `unset_if_exists` | `unset "$1"` | no internal locals — safe |
| `src/env.sh:298-299` | `${!key}` | iterates hardcoded `BASHUNIT_*` names — no caller-provided names, safe |
| `src/assert*.sh`, `src/main.sh` | `eval "$cmd"` | command-string execution, not outvar pattern — safe |

`make sa` clean, `make lint` clean, unit suite 848 passed.

Chemaclass:
Switched to simpler option after review.

**runner.sh hot-path helpers** now return via dedicated global slots, no eval, no prefix:
- `_BASHUNIT_RUNNER_FIELD_OUT` (`extract_encoded_field`)
- `_BASHUNIT_RUNNER_TOTAL_OUT` (`compute_total_assertions`)
- `_BASHUNIT_RUNNER_TYPE_OUT` (`extract_subshell_type`)
- `_BASHUNIT_RUNNER_OUTPUT_OUT` (`format_subshell_output`)

Internal locals back to natural names (`input`, `val`, `subshell_output`, etc.). Same perf as the eval-outvar version (no `$(...)` fork) and the dynamic-scope hazard is structurally gone — helper never touches caller-named variables.

**test_doubles.sh** — the real fix was on the **constructed global names**, not on helper locals. Internal locals restored to natural names; spy state moved from `${variable}_times_file` to `_BASHUNIT_SPY_${variable}_TIMES_FILE`. A caller doing `local foo_times_file=...` can't collide with `_BASHUNIT_SPY_foo_TIMES_FILE`.

**Rule in bash-style.md** rewritten as three patterns in priority order:
1. Preferred: dedicated global return slot (the new runner pattern + the existing `_BASHUNIT_BRANCH_ARMS_OUT` in coverage.sh)
2. For helpers that build dynamic names: namespace the constructed name (`_BASHUNIT_SPY_*`)
3. Last-resort outvar-by-name with `__bu_` prefix on locals + regression test (none in tree right now)

Issue #722, opened 2026-06-16
Line continuations not counted as covered in report


RevanProdigalKnight:
| Q                | A                        |
|------------------|--------------------------|
| OS               | Linux (Cygwin on Windows 10) |
| Shell & version  | bash 5.2.21(1)-release |
| bashunit version | 0.39.1                    |

#### Summary

Lines preceded by a line ending with a backslash (i.e.: splitting long commands across multiple lines) are not counted as covered in the coverage report

#### How to reproduce

```sh
function some_function() {
  echo "hello world" |
    SHELL_VAR=1 \ # <-- this line is counted as covered
    grep 'hello' # <-- this line is not counted as covered
}
```

#### Expected behavior

All lines following a line ending in a backslash should be counted as covered

Chemaclass:
Fixed in #723 (merged to main). Backslash line-continuation lines now inherit their statement's hit count. See src/coverage.sh:652-714.

PR #723, opened 2026-06-16
fix(coverage): count backslash line-continuation lines as covered

truffle-dev:
## Background

Related #722

Bash's DEBUG trap attributes a multi-line statement's execution to the line where the statement starts, so the lines after a trailing `\` never receive their own hit and are reported as uncovered. The issue's example shows the line with the backslash counted, but the following line missed.

## Changes

- Propagate each start line's hit count forward across its continuation chain in `get_all_line_hits`. This is the single function feeding the percentage, lcov, html and uncovered-line reports, so the fix lands everywhere at once.
- Add `bashunit::coverage::_ends_with_continuation`, which detects a trailing unescaped backslash (odd number of trailing backslashes, no trailing whitespace, not a comment line).
- Unit tests for `_ends_with_continuation` (single/odd/even backslashes, trailing whitespace, comments, empty line) and for the forward propagation in `get_all_line_hits`.

Reproducing the issue's snippet under a real coverage run, before vs after:

```
# before
src/calc.sh    2/3 lines ( 66%)
# after
src/calc.sh    3/3 lines (100%)
```

`make test` (868 passed, 3 pre-existing skips), `shellcheck -xC` clean on both changed files. Bash 3.0 compatible (indexed arrays only).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fixed
+- Coverage report now counts backslash line-continuation lines as covered: a multi-line statement's hit is propagated forward across its continuation chain, so the lines after a trailing `\` are no longer reported as uncovered (#722)
+
 ### Changed
 - Documentation and project URLs now point to the new primary domain `bashunit.com` (old `bashunit.typeddevs.com` continues to work as a redirect)
 - The docs site now deploys to GitHub Pages on the `bashunit.com` custom domain (`deploy-gh-pages.yml`); the installer is copied into the site root from the repo's canonical `install.sh`
--- a/src/coverage.sh
+++ b/src/coverage.sh
@@ -647,8 +647,27 @@ function bashunit::coverage::compute_file_coverage() {
   echo "${executable}:${hit}"
 }
 
+# Detect whether a source line ends with a Bash line-continuation, i.e. an
+# odd number of unescaped trailing backslashes with no trailing whitespace.
+# Comment lines never continue. Used to propagate coverage hits from a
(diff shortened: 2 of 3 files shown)


Chemaclass (approved):
Approve. Verified empirically — reproduced issue #722's exact snippet under a real coverage run, PR branch vs main:

```
main:  src/calc.sh  2/3 lines (66%)   DA:3,1 DA:4,1 DA:5,0   # grep line uncovered
PR:    src/calc.sh  3/3 lines (100%)  DA:3,1 DA:4,1 DA:5,1   # grep line covered
```

Line after the backslash continuation flips uncovered→covered. Matches the issue exactly.

Code review:
- Algorithm traced and correct. The `h > 0` guard before setting `carry` is the key safety property — prevents false propagation into heredoc bodies / data lines that end with `\` but have no hit.
- `_ends_with_continuation` handles single/odd/even backslashes, trailing whitespace, comments, empty lines. Well tested.
- `< <(grep …)` process substitution is the right fix so the `counts` array survives (a pipe would lose it in a subshell).
- Pre-increment `((++_i))` avoids the set -e post-increment trap. Bash 3.0 compatible (indexed arrays only).

Minor / non-blocking:
- Output ordering changed from lexical (`sort | uniq -c`) to numeric (1..total). Arguably an improvement; lcov/html consumers build maps so ordering shouldn't matter.
- Adds a full source-file read per file even when no continuations exist. Negligible on the coverage path.
- Optional: add a test asserting the `h > 0` guard — a continuation-looking line with zero hits should NOT get a phantom hit (currently implied, not explicitly tested).

PR #913, opened 2026-07-28
fix: Test failures on Bash 5.3 macOS (nix-shell)

### LINKED ISSUE #912: 7 test failures on Bash 5.3 macOS (nix-shell)
| Q                | A                                      |
|------------------|----------------------------------------|
| OS               | macOS (aarch64-apple-darwin25.3.0)     |
| Shell & version  | bash 5.3.9(1)-release (nix-shell)      |
| bashunit version | 0.43.0                                 |

#### Summary

7 test failures under `nix-shell --pure` on macOS with Bash 5.3. CI passes because macOS runners use Apple's Bash 3.2. Three bugs:

1. `render_diff` calls `git diff` without `--no-ext-diff`. This breaks for anyone with `diff.external` configured, e.g. I have difftastic configured. (2 failures)
2. `shell_time()` uses `LC_ALL=C` inside `$()`. This results in a SIGSEGV on Bash 5.3 macOS (5 failures)
3. Syntax error test uses `LC_ALL=C` inside `$()`. Same SIGSEGV (1 failure)

Bugs 2 and 3 share a root cause: Bash 5.3 on macOS segfaults when `LC_ALL=C` or `LANG=C` is used as a temporary env prefix inside a command substitution. Minimal reproducer:

```bash
# Bash 5.3.9(1)-release, aarch64-apple-darwin25.3.0
bash -c 'x=$(LC_ALL=C echo hello); echo exit=$?'
# → exit=139 (SIGSEGV)
```

#### Current behavior

```
Tests:      7 failed, 1515 total
```

#### How to reproduce

```bash
nix-shell --pure --run "./bashunit --simple --parallel"
```

Requires the `shell.nix` in the repo root. The key ingredient is Bash 5.3 on macOS. Any setup that provides that will hit bugs 2 and 3. Bug 1 requires `diff.external` in git config (e.g. difftastic).

#### Expected behavior

All tests pass.

tricktron:
## Background

Closes #912.

I first stumbled on this issue in https://github.com/NixOS/nixpkgs/pull/543029.

If merged we can update and simplify the bashunit derivation in nixpkgs and all tests will pass even in the sandbox on darwin.

## Changes

- `render_diff` didn't bypass external diff tools. Anyone with e.g. difftastic got broken output (2 failures)
- `shell_time()` and one acceptance test used `LC_ALL=C` inside `$()`, which segfaults on Bash 5.3 macOS. I removed since it was unnecessary in both places (5 failures).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@
 - An unreadable or truncated parallel `.result` file counts as a failed test instead of aborting aggregation
 - `release.sh` reports a failed rollback as failed
 - Docs: the quickstart's duration format, and the real `BASHUNIT_SHOW_EXECUTION_TIME` default (`auto`)
+- Word-diff output was broken for users with `diff.external` configured (e.g. difftastic); `render_diff` now passes `--no-ext-diff` to `git diff`
+- `shell_time()` crashed (SIGSEGV) on Bash 5.3 macOS due to `LC_ALL=C` inside a command substitution; removed since callers already handle both decimal separators
 
 ### Removed
 - `bin/create-pr`, an unreferenced vendored copy of [Chemaclass/create-pr](https://github.com/Chemaclass/create-pr); use the upstream tool (#867)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 - The `--parallel` unsupported-OS warning no longer claims Alpine is excluded
 
 ### Fixed
+- A configured external git differ (`diff.external` / `GIT_EXTERNAL_DIFF`, e.g. difftastic) no longer blanks the multiline and snapshot failure diffs, which now render with `--no-ext-diff` (#912)
+- Time reads and the JUnit report no longer set the locale with a temporary-environment prefix (`LC_ALL=C cmd`), which segfaults inside a command substitution on Bash 5.3.9 macOS (#912)
 - Call assertions (`assert_not_called`, `assert_have_been_called*`) fail with `was never registered as a spy` instead of reporting zero calls when the name was never spied — a typo used to pass silently (#895)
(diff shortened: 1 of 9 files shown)


Chemaclass (approved):
Nice catch, and thanks for the clear reproducer in #912. That made both bugs easy to confirm. You got here first, so I'm merging yours and closing my #914.

I pushed a few small things on top so you don't have to round-trip:

- `src/reports.sh` had the same `LC_ALL=C` prefix in its two JUnit time reads, so `--report-junit` would still crash on Bash 5.3. There the `C` locale is actually doing something (it keeps awk printing `1.234` and not `1,234`), so I used `env LC_ALL=C awk` instead of removing it.
- A test for each fix. CI can't catch these on its own: no runner has `diff.external` set, and none runs Bash 5.3 on macOS. Both tests fail without your changes.
- A short comment next to each removal so nobody puts the prefix back, plus the CHANGELOG entries moved to the top of Fixed with the `(#912)` link.

Dropping `LC_ALL=C LANG=C` from the acceptance test was the right call. Those assertions only look at bashunit's own output, and the Spanish/Brazilian/Japanese jobs run the acceptance tests too, so they'd have told us otherwise.

All green here on macOS Bash 3.2 (`make test`, `--parallel`, `make sa`, `make lint`). Thanks again!

tricktron:
@Chemaclass Wow, you are fast. Thank you. Let me know when you have cut a new release with tag, then I start bumping and simplifying the nix derivation.

Chemaclass:
@tricktron new release out! https://github.com/TypedDevs/bashunit/releases/tag/0.44.0

tricktron:
Thanks @Chemaclass. I opened the Nixpkgs PR in https://github.com/NixOS/nixpkgs/pull/547099

PR #914, opened 2026-07-28
fix(compat): ignore an external git differ, drop locale env prefixes

Chemaclass:
## 🤔 Background

Related #912

Seven tests failed on macOS with Bash 5.3 under `nix-shell --pure`, and two more for anyone with an external git differ configured. CI never sees either: macOS runners ship Apple's Bash 3.2, and no runner has `diff.external` set.

## 💡 Changes

- The failure/snapshot diff renderer now passes `--no-ext-diff`, so a configured `diff.external` (difftastic) no longer replaces the word-diff with empty output.
- Time reads and the JUnit report no longer set the locale with a temporary-environment prefix (`LC_ALL=C cmd`), reported to segfault inside a command substitution on Bash 5.3.9 macOS. `env LC_ALL=C cmd` keeps the C radix; the clock never needed the prefix, since the value is expanded before the temporary environment applies.
- The syntax-error acceptance test drops the same prefix form.
- A source-level gate keeps the prefix out of `src/`, since no CI job runs that bash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 - The `--parallel` unsupported-OS warning no longer claims Alpine is excluded
 
 ### Fixed
+- A configured external git differ (`diff.external` / `GIT_EXTERNAL_DIFF`, e.g. difftastic) no longer blanks the multiline and snapshot failure diffs, which now render with `--no-ext-diff` (#912)
+- Time reads and the JUnit report no longer set the locale with a temporary-environment prefix (`LC_ALL=C cmd`), reported to segfault inside a command substitution on Bash 5.3.9 macOS (#912)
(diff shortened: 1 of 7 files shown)


Chemaclass:
Closing in favour of #913, which reported and fixed the same three bugs first.

Everything this PR had that #913 was missing is now on that branch: the `env LC_ALL=C awk` fix in `src/reports.sh`, the external-differ regression test, and the `bash_compatibility_test.sh` gate against the crashing locale prefix. Nothing is lost here.

#913 also carries two fixes this PR did not have: the init acceptance test writes its log under `$TMP_DIR` instead of `/tmp`, and the osx-with-perl clock test pins `_BASHUNIT_CLOCK_NOW_IMPL` so it keeps testing the perl path on Bash 5.

Issue #929, opened 2026-07-31
bug: `bashunit doc` fails on missing default bootstrap instead of showing docs


objctp:
| Q                | A                        |
|------------------|--------------------------|
| OS               | macOS (also reproducible on Linux) |
| Shell & version  | bash 3.2 / 5.2           |
| bashunit version | 0.44.0                   |

#### Summary

`bashunit doc` fails with `Error: cannot read the bootstrap file: 'tests/bootstrap.sh'.` when no bootstrap file exists, instead of displaying the assertion documentation.

#### Current behavior

Running `bashunit doc` (the plain form, with no `--custom` and no `--boot`/`-e`/`--env`) in a project that does not yet have a `tests/bootstrap.sh` file prints an error and exits with code 1:

$ bashunit doc
Error: cannot read the bootstrap file: 'tests/bootstrap.sh'.

This happens even though the bootstrap is only needed to discover *custom* assertions. The default `BASHUNIT_BOOTSTRAP` value (`tests/bootstrap.sh`) is treated as mandatory before any documentation is printed.

#### How to reproduce

```bash
# Start from a clean project (no bootstrap file)
mkdir my-project && cd my-project
# (no tests/bootstrap.sh present)

# Plain documentation listing, as shown in the official docs:
$ bashunit doc [assert | contain]
# -> Error: cannot read the bootstrap file: 'tests/bootstrap.sh'.
#    exit code 1, no documentation shown
```
The same error also appears for bashunit doc --custom when no bootstrap exists (it should instead report "No custom assertions found.").

#### Expected behavior
Per the official doc documentation (https://bashunit.com/command-line#doc), bashunit doc "Displays documentation for assertion functions" and lists all assertions by default — no bootstrap file should be required.
- bashunit doc → prints all assertion documentation, regardless of whether a bootstrap file exists.
- bashunit doc --custom → prints "No custom assertions found." when no bootstrap/custom assertions are present (no hard error).
- Only an explicitly named bootstrap (bashunit doc --boot <file> / -e / --env) should error when the file cannot be read.

Chemaclass:
Done in #930 (merged).

Issue #875, opened 2026-07-25
A missing --env bootstrap or an unwritable report path fails silently and exits 0


Chemaclass:
## 🤔 Background

Same class as #871 and #873: an unusable path is accepted, the work it was meant to do is skipped, and the run still reports success.

## `--env` / `--boot` with a missing file

`src/main.sh` sources the bootstrap without checking that it exists. A typo'd path leaks a raw shell error, **runs no tests at all**, and exits `0`:

```console
$ ./bashunit --env missing.sh tests/
src/main.sh: line 242: missing.sh: No such file or directory
$ echo $?
0
```

That is the entire output — no header, no test lines, no totals. With a valid bootstrap the same command runs the suite normally, so the failure mode is "green build that tested nothing".

## Report paths that cannot be written

`--report-json`, `--log-junit` and friends do not check that the destination is writable. The suite runs, the report is silently not produced, and the run exits `0`:

```console
$ ./bashunit --report-json /nope/dir/o.json tests/
 All tests passed
src/reports.sh: line 207: /nope/dir/o.json: No such file or directory
$ echo $?
0
$ test -f /nope/dir/o.json || echo "no report written"
no report written
```

A CI job that runs tests and then uploads the report gets a green build and an empty artifact.

## Also unvalidated

- `--seed abc` is silently ignored (exit `0`). Lower impact than the above, but the same gap — #873 fixed the other numeric options.

## Suggested fix

Fail fast at the boundary, matching `bashunit::main::validate_config_or_exit` added in #874:

- `--env`/`--boot`: error and exit non-zero when the file does not exist or is not readable.
- Report options: verify the parent directory exists and is writable before the run starts, so the failure is reported up front rather than after the suite has passed.
- `--seed`: validate as a non-negative integer.

Chemaclass:
Done in #876 (merged).

PR #930, opened 2026-07-31
fix(doc): don't error on missing default bootstrap file (#929)

objctp:
Plain 'bashunit doc' now lists all assertions without requiring a bootstrap; only an explicitly named --boot/-e/--env file still errors when unreadable.

## Background

Related #929

`bashunit doc` treated the default `BASHUNIT_BOOTSTRAP` (`tests/bootstrap.sh`) as mandatory and aborted with `Error: cannot read the bootstrap file: 'tests/bootstrap.sh'.` before printing anything. This contradicts the documented behaviour, where plain `bashunit doc` lists all assertions without requiring a bootstrap file.

## Changes

- Make the default bootstrap optional in `bashunit::main::cmd_doc`: a missing *default* bootstrap no longer errors. Plain `bashunit doc` prints all assertions, and `bashunit doc --custom` prints "No custom assertions found." when none exist.
- An explicitly named bootstrap (`--boot` / `-e` / `--env`) still errors when unreadable, preserving intended behaviour.
- Added a `CHANGELOG.md` entry under Unreleased → Fixed.

## Checklist

- [ ] I updated the documentation to reflect the changes (no change needed — behaviour now matches the existing `doc` docs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - `build.sh` dedupes embedded files by repo-relative path. The previous basename key compared the top-level loop's relative paths against the recursion's absolute ones, so a file reached from two places could be bundled twice in the released binary; it also collided for same-named files in different directories (#923)
+- `bashunit doc` no longer errors when the default bootstrap file is missing (#929)
 
 ## [0.44.0](https://github.com/TypedDevs/bashunit/compare/0.43.0...0.44.0) - 2026-07-29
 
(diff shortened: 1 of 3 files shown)


Chemaclass (approved):
Fix is correct and now covered.

Verified independently: with the fix, a missing *default* bootstrap prints the catalogue and exits 0; an explicitly named `--boot` that is unreadable still exits 1 with the error. This matches what `cmd_test` already does — explicit `--env` hard-errors (#875), the default is `[ -f ] && source`.

The review blocker was the missing test; added in 6c9cbbd and verified to fail against the pre-fix `src/main.sh` and pass with it.

Issue #1027, opened 2026-08-09
feat(assert): assert_command_available


Chemaclass:
## Problem

`assert_command_not_found` exists (`src/assert/core.sh`) but has no positive counterpart. Asserting that a dependency **is** installed — the first thing an installer test, a Docker image test or a bootstrap test wants to check — has to go through the helper:

```bash
assert_true "bashunit::is_command_available jq"
```

which reports the failure as a boolean assertion over an opaque string rather than naming the missing command. The check itself already exists as `bashunit::is_command_available` (`src/api/globals.sh`); only the assertion wrapper is missing.

Every other assertion family in bashunit ships both directions.

## Proposal

```bash
assert_command_available jq
```

- Passes when the command resolves in `PATH` (builtins and functions included, consistent with `bashunit::is_command_available`).
- Fails with a message naming the command, in the same voice as `assert_command_not_found`.
- Implemented on top of `bashunit::is_command_available` so there is one definition of "available", not two.

Also consider aliasing to match whichever name reads better next to the existing `assert_command_not_found` — but ship exactly one name, not two.

## Acceptance criteria

- [ ] `assert_command_available bash` passes
- [ ] `assert_command_available definitely_not_a_command` fails with the command name in the message
- [ ] Behaviour for shell builtins and shell functions matches `bashunit::is_command_available` and is documented
- [ ] A missing argument is a usage error (#983 precedent)
- [ ] Symmetric with `assert_command_not_found`: the two never both pass or both fail for the same input
- [ ] Implemented via `bashunit::is_command_available`, no second definition of availability

## Repo checklist (agent)

- **TDD**: RED → GREEN → REFACTOR. Use `/add-assertion` if helpful.
- **Bash 3.0+ only**: no `printf -v`, no `+=`, no `declare -A`, no `[[ ]]`, no `${var,,}`, no `&>>`.
- Assertions are **bare-named** (`assert_x`), helpers are `bashunit::`-namespaced. Start every assertion with `bashunit::assert::should_skip && return 0` and report through `bashunit::assertion_failed` / `bashunit::assertion_passed`, matching its neighbours in the same file.
- A missing required argument must go through `bashunit::assert::usage_error`, not compare against an empty string (#983).
- **Tests**: mirror the src layout — `tests/unit/assert/<file>_test.sh`. There is no `assert_fails`; test failure output by comparing against `print_failed_test`, as the existing tests do.
[...]

Issue #1008, opened 2026-08-09
feat(tags): tag expressions (AND / NOT) and file-level @tags


Chemaclass:
## Problem

Tag selection is weaker than the systems people migrate from.

- `--tag` is repeatable but **OR-only** (`bashunit test --help`: "repeatable, OR logic"). There is no way to say "slow **and** db".
- There is no negation inside an expression; `--exclude-tag` is a separate flag that always wins.
- Tags are **per function only**. `src/helper/tags.sh:10` `build_tags_map` scans the contiguous comment block directly above each `function` definition, so tagging a whole file means repeating `# @tag integration` above every test in it.

bats-core supports `--filter-tags 'a,!b'` with AND semantics and has both `# bats test_tags=` and `# bats file_tags=`.

## Proposal

**File-level tags** — a `# @tags <list>` comment before the first function definition (or anywhere at top level) applies to every test in that file, unioned with per-function `# @tag`.

**Tag expressions** in `--tag`:

```
--tag slow              # has slow                     (unchanged)
--tag slow --tag db     # slow OR db                   (unchanged, back-compat)
--tag 'slow&&db'        # slow AND db
--tag '!slow'           # NOT slow
--tag 'db&&!slow'       # db AND NOT slow
```

Precedence stays as documented: `--exclude-tag` still wins over any `--tag` match.

## Where to change

- `src/helper/tags.sh:10` `build_tags_map` — single awk pass; extend it to also collect file-level tags and union them in.
- `src/helper/tags.sh:99` `function_matches_tags` — the expression evaluator lives here.
- `src/helper/tags.sh:74` `tags_for_function`.

## Acceptance criteria

- [ ] `# @tags a b` above the first function applies both tags to every test in the file
- [ ] File-level and function-level tags union (no override, no duplicate match)
- [ ] `--tag 'a&&b'` selects only tests carrying both
- [ ] `--tag '!a'` selects only tests without `a`
- [ ] `--tag 'a&&!b'` works
- [ ] Repeated `--tag` flags keep OR semantics between them (no behaviour change for existing users)
- [ ] `--exclude-tag` still wins over an expression match
- [ ] A malformed expression (`'a&&'`, `'&&'`, `'!'`) exits non-zero with a usage error, not a silent full-suite run (#871 class of bug)
- [ ] The memoization in `build_tags_map` still holds — no per-function rescan (#773)
[...]

Issue #1007, opened 2026-08-09
feat(cli): --list / --dry-run to show which tests would run


Chemaclass:
## Problem

There is no way to ask bashunit *which tests it would run*. Every selection mechanism — `--filter`, `--tag`, `--exclude-tag`, `--shard i/n`, `--random-order --seed`, `--rerun-failed`, `file::fn`, `file:LINE` — is only observable by actually running the suite.

Consequences today:

- `--shard 1/4` cannot be checked for balance without four full runs.
- A typo'd `--filter` silently runs 0 tests and exits 1 with "No tests found" (`src/console/summary.sh:128`), indistinguishable from a genuinely empty suite.
- Editor/CI tooling has no machine-readable list of test ids.

bats-core has `-c/--count`, shellspec has `--dry-run` and `--list examples`.

## Proposal

```
--list                    Print the tests that would run, then exit 0
--list-format <fmt>       text (default) | json
--dry-run                 Alias of --list (shellspec-compatible spelling)
```

- Honours **every** selection flag, applied in the same order as a real run, including shard slicing and `--random-order --seed` ordering.
- `text` output: one `path/to/file_test.sh::test_function_name` per line, followed by `N tests` on stderr.
- `json` output: `{"count": N, "tests": [{"file": "...", "function": "...", "name": "human readable", "line": 12, "tags": ["slow"]}]}`.
- No test body is executed, no bootstrap side effects beyond sourcing, no reports are written.
- Exit code `0` even when the list is empty (this is a query, not a run).

## Where to change

- `src/helper/discovery.sh:62` `get_functions_to_run` and `:134` `find_total_tests` already do the selection work — the list command should reuse them rather than re-implement filtering.
- `src/helper/tags.sh:99` `function_matches_tags` for the tag column.
- `src/helper/discovery.sh:243` `get_function_line_number` for the line column.
- `src/main/run.sh` — branch before the runner loop.
- `src/helper/naming.sh` `normalize_test_function_name` for the human-readable name.

## Acceptance criteria

- [ ] `bashunit --list tests/` lists every test, one per line, and exits 0
- [ ] `--list` respects `--filter`, `--tag`, `--exclude-tag`, `--shard`, `--rerun-failed`, `file::fn` and `file:LINE`
- [ ] `--list --random-order --seed 42` prints the same order that `--random-order --seed 42` would execute
[...]

Issue #983, opened 2026-08-06
feat(assert): a missing argument reports a failed assertion, not a usage error


Chemaclass:
## Summary

Call a two-argument assertion with one argument and it does not complain. It compares against the empty string and reports an ordinary failure:

```bash
function test_wrong_arg_count() {
  assert_same "only-one"
}
```

```
✗ Failed: Wrong arg count
    Expected 'only-one'
    but got  ''
```

That output is indistinguishable from a genuine failure where the code under test really did produce an empty string. The reader debugs their code; the bug is in the test.

## Why this is the expensive kind of mistake

Empty is the single most common *real* value in shell — an unset variable, a command that printed nothing, a failed capture. So the false story this tells is also the most plausible one. Someone hitting it will go and check why their function returned nothing, because that is exactly what the message says happened.

It is also easy to reach: `assert_same "$expected"` after deleting the second argument during a refactor, or a typo'd `"$actual"` that expands to nothing and silently collapses the argument count.

Related and worth handling together: argument order in this catalogue is not uniform, so a swapped pair is another way to produce a confusing-but-plausible failure. `assert_same` and `assert_contains` take expected/needle first; `assert_json_contains` takes `key expected json` with the subject **last**. I hit that myself while researching this — `assert_json_contains '{"a":[1,2,3]}' '.a|length' "3"` reports `Expected '.a | length' but got '{…'`, which is the assertion faithfully comparing the two arguments it was handed. Nothing is wrong except that no one can tell.

## Proposal

**Arity checking** on the assertion entry points. When too few arguments are supplied, fail with a usage error rather than a comparison:

```
✗ Error: Wrong arg count
    assert_same expects 2 arguments (expected, actual), got 1
```

Design points worth settling before implementing:

* **Report as an error, not a failure.** This is a defect in the test, the same class as calling an assertion that does not exist — which this framework already surfaces as `Error`, not `Failed`. Reusing that channel keeps the distinction the reader needs.
* **Only under-supply, not over-supply.** Several assertions take a trailing optional argument (a label override, an `nth` index), so a strict upper bound would be wrong or would need per-assertion tables. Under-supply is unambiguous and covers the real failure mode.
* **Do not use `$#` naively** where an assertion legitimately accepts an empty string as its second argument — `assert_empty "$x"` and `assert_same "" "$x"` are valid. The check is on *count supplied*, which `$#` gives correctly, so this works; it just has to be `$#` and never `[ -z "$2" ]`.

[...]

PR #1033, opened 2026-08-09
feat(assert): add assert_command_available

fzlzjerry:
## Background

Related #1027

bashunit already exposes `bashunit::is_command_available`, but users could not assert that a required command resolves with a diagnostic that names the missing command. The workaround — `assert_true "bashunit::is_command_available jq"` — reported the failure as a boolean over an opaque string.

## Changes

- add `assert_command_available <command>` on top of the existing availability helper, so there is one definition of "available". A missing argument is a usage error (#983 precedent); an optional second argument overrides the failure label, matching the rest of `src/assert/core.sh`
- cover external commands, shell builtins, shell functions, the missing-command message, the custom label, the assertion counters, and the inverse relationship with `assert_command_not_found` in both directions
- generalise `tests/unit/assert/arity_test.sh`: it now builds the "one argument short" call from the declared arity instead of branching on it, so the 1-argument case needed no new branch
- add the assertion to the documentation, the generated `bashunit doc` snapshot, both completion scripts, the changelog, the agent skill catalogue and the assertion counts

## Maintainer notes

Rebased onto `main` (conflict was the `## Unreleased` changelog block against #1007, #1008 and #1035) and refactored on top of the contributor's commits:

- trimmed the doc block to the house `Arguments:` form — the note about calling `bashunit::assert::fail_with` instead of the `bashunit::assertion_failed` facade repeated `fail_with`'s own documented contract
- dropped the added `tests/acceptance/bashunit_hook_failure_test.sh` case. It spent a full CLI run pinning the facade's frame depth for an assertion that never routes through the facade, and the two existing guards in that file already pin `fail_with`'s depth generically
- added the symmetry test the issue asks for (`assert_command_available` and `assert_command_not_found` never agree about the same command)

## Validation

- `./bashunit tests/` — 1786 passed, 13 skipped, 4 incomplete, 2 snapshot
- `./bashunit --parallel tests/` — 1747 passed, 11 skipped, 4 incomplete, 2 snapshot
- `/bin/bash ./bashunit --strict tests/unit/assert/arity_test.sh` on Bash 3.2 — exercises the empty-array `set -u` guard in the new arity loop
- `make sa` and `make lint` — OK
- `./build.sh` + `tests/acceptance/bashunit_test.sh` — built binary docs match dev docs
- mutation check: relaxing the new arity guard to `-lt 0` turns the arity test red, so it is not vacuous

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - `--list` (alias `--dry-run`) prints the tests a run would execute, without running them; `--list-format json` emits file, function, name, line and tags. Honours every selection flag, including `--shard` and `--random-order --seed` ordering (#1007)
 - `# @tags a b` above any top-level line applies those tags to every test in the file, unioned with per-function `# @tag` (#1008)
 - `--tag` accepts expressions: `'a&&b'` (AND) and `'!a'` (NOT), combinable as `'a&&!b'`. Repeated `--tag` flags keep OR semantics, and `--exclude-tag` still wins (#1008)
+- `assert_command_available <command>` asserts that an external command, shell builtin or function resolves through `command -v` (#1027)
 - The coverage engine in use is reported by `--verbose`, and an explicit `BASHUNIT_COVERAGE_ENGINE=xtrace` that the running Bash cannot honour now warns instead of being silently ignored (#1005)
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@
 ## Why bashunit
 
 A lightweight, fast testing framework for **Bash 3.0+**, focused on developer experience.
-It ships 73 assertions plus spies, mocks, data providers, snapshots and more.
+It ships 75 assertions plus spies, mocks, data providers, snapshots and more.
 
 ## Quick start
 
--- a/completions/_bashunit
+++ b/completions/_bashunit
@@ -17,7 +17,7 @@ _bashunit() {
     assert_fns=(
       assert_array_contains assert_array_length assert_array_not_contains
       assert_arrays_equal assert_assertion_fails assert_assertion_fails_with
-      assert_assertion_passes assert_command_not_found assert_contains
+      assert_assertion_passes assert_command_available assert_command_not_found assert_contains
       assert_contains_ignore_case assert_date_after assert_date_before
       assert_date_equals assert_date_within_delta assert_date_within_range
       assert_directory_exists assert_directory_not_exists assert_duration
--- a/completions/bashunit.bash
+++ b/completions/bashunit.bash
@@ -30,7 +30,7 @@ _BASHUNIT_COMPLETIONS_TEST_OPTS="--assert --boot --coverage --coverage-exclude \
 
 _BASHUNIT_COMPLETIONS_ASSERT_FNS="assert_array_contains assert_array_length \
 assert_array_not_contains assert_arrays_equal assert_assertion_fails \
-assert_assertion_fails_with assert_assertion_passes assert_command_not_found \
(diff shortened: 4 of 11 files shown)


Chemaclass:
Reviewed and pushed two follow-up commits to this branch.

**Blocker fixed (f03f8db).** `assert_command_available` reported through `bashunit::assertion_failed`, the public facade meant for user-defined custom assertions. That facade adds a stack frame on top of `bashunit::assert::fail_with`, whose label fallback (`fallback_depth 3`) compensates for exactly one. With no `test_*` frame on the stack the label resolved to the facade:

```
✗ Failed: Bashunit::assertion failed        # before, failing inside set_up
✗ Failed: Assert command available          # after
```

Same class of regression already pinned for `bashunit::assert::label` in `tests/acceptance/bashunit_hook_failure_test.sh`; added a case there for the facade frame too. Built-in assertions must call `bashunit::assert::fail_with` / `bashunit::state::add_assertions_passed` directly, like every other assertion in `src/assert/core.sh`.

Also in that commit:
- dropped the trailing space in `"to be available but was "` — `print_failed_test` already emits one, so the output read `to be available but was  'not found'`
- accepted the conventional label override as `$2` (matches `assert_empty` / `assert_not_empty`), with a test
- replaced `test_assert_command_available_is_symmetric_with_assert_command_not_found`: it executed the command and re-asserted that a missing one exits 127, which is `assert_command_not_found`'s contract, not a symmetry between the two assertions (they aren't inverses — one checks resolution, the other an exit code). Swapped for a counter-movement test
- reworded the docs link so `bashunit doc` renders it like the other cross-references, and regenerated the snapshot

**Docs (01f3442).** Added the name to the "Exit codes" bullet in `docs/public/bashunit-skill.md` — the count was bumped there but the list is what keeps an agent from inventing assertion names.

The `73 → 75` bump is correct, by the way, not an off-by-one: `main` was stale by two. 75 non-spy assertion functions; the 7 spy assertions are the "plus spies".

Local gate green: full suite, `--parallel`, `make sa`, `make lint`.

Chemaclass (approved):
Reviewed after rebasing onto `main` and refactoring on top of the original commits.

**Blockers:** none.

**Checked**

- Bash 3.0+: no `declare -A`, `[[ ]]`, `${var,,}`, negative indexing or `&>>`. The one new array (`args` in `arity_test.sh`) is appended with `arr[${#arr[@]}]=` and expanded as `"${args[@]+"${args[@]}"}"`, so the empty case does not trip `set -u` on Bash 3.x — verified with `/bin/bash ./bashunit --strict` on 3.2.57.
- Availability has one definition: the assertion delegates to `bashunit::is_command_available`, so builtins and shell functions behave identically to the helper, and the command is resolved rather than executed.
- Failure output follows the `fail_with` voice of the neighbouring file/folder assertions and names the command. Custom label goes in `$2`, matching the rest of `src/assert/core.sh`.
- Missing argument is a usage error (exit 2), per the #983 precedent, and the arity provider covers it. Mutation check: relaxing the guard to `-lt 0` turns the arity test red, so the case is real.
- Docs, `bashunit doc` snapshot, both completion scripts, the agent skill catalogue, the assertion counts and the changelog are all updated. `./build.sh` plus `tests/acceptance/bashunit_test.sh` confirm the built binary's docs match dev docs.
- `make sa` and `make lint` are clean; full and `--parallel` suites pass locally.

**Nit, not acted on:** `Expected 'jq' / to be available but 'not found'` reads a little stiff, but it is the same shape as `to exist but 'do not exist'` in `src/assert/files.sh`, so it stays consistent with the family.

Approving; merging once every CI job is green.

PR #1037, opened 2026-08-10
feat(assert): add numeric range assertions

### LINKED ISSUE #1026: feat(assert): assert_between / assert_not_between for numeric ranges
## Problem

Numeric comparison covers the open-ended cases — `assert_greater_than`, `assert_greater_or_equal_than`, `assert_less_than`, `assert_less_or_equal_than`, `assert_within_delta` — but not the bounded one. Asserting "the duration was between 100 and 500 ms" or "the exit code is in the 4xx range" takes two assertions, which reports two results and two failure messages for one logical check.

Dates already have the bounded form: `assert_date_within_range` (`src/assert/dates.sh`). Numbers do not.

## Proposal

```bash
assert_between <min> <max> <actual> [label]
assert_not_between <min> <max> <actual> [label]
```

- Inclusive bounds, matching `assert_date_within_range` — check that assertion first and mirror its semantics and argument order rather than choosing independently.
- Supports integers and decimals, reusing the fixed-point path added for `assert_within_delta` in #979 (with the `bc`/`awk` fallback) instead of forking `bc` per call.
- Reports **once**, not twice — use `bashunit::assert_once` (`src/assert/once.sh`) if implemented in terms of other assertions, so the count matches one assertion (#917).

## Acceptance criteria

- [ ] `assert_between 1 10 5` passes; `assert_between 1 10 11` fails
- [ ] Bounds are inclusive: `assert_between 1 10 1` and `assert_between 1 10 10` both pass
- [ ] Decimals work: `assert_between 0.1 0.3 0.2`
- [ ] Negative numbers work
- [ ] `min > max` is a usage error, not a silently always-failing assertion
- [ ] Non-numeric input is a usage error (#983 / #984 precedent), not a comparison against empty
- [ ] The failure message shows all three values and which bound was violated
- [ ] Counts as exactly one assertion (#917)
- [ ] `assert_not_between` is the exact negation and is tested independently
- [ ] No `bc` fork on the common integer path (#979 precedent) — the fork-budget acceptance tests must still pass

## Repo checklist (agent)

- **TDD**: RED → GREEN → REFACTOR. Use `/add-assertion` if helpful.
- **Bash 3.0+ only**: no `printf -v`, no `+=`, no `declare -A`, no `[[ ]]`, no `${var,,}`, no `&>>`.
- Assertions are **bare-named** (`assert_x`), helpers are `bashunit::`-namespaced. Start every assertion with `bashunit::assert::should_skip && return 0` and report through `bashunit::assertion_failed` / `bashunit::assertion_passed`, matching its neighbours in the same file.
- A missing required argument must go through `bashunit::assert::usage_error`, not compare against an empty string (#983).
- **Tests**: mirror the src layout — `tests/unit/assert/<file>_test.sh`. There is no `assert_fails`; test failure output by comparing against `print_failed_test`, as the existing tests do.
- **Docs**: add the entry to `docs/assertions.md`, then **regenerate the `bashunit doc` acceptance snapshot** (editing that file breaks it).
- Update the assertion count in `README.md` ("It ships 73 assertions…" — currently already stale at 74) and add a `CHANGELOG.md` line under `## Unreleased`.
- **Gates**: `make sa`, `make lint`, `./bashunit tests/`, `./bashunit --parallel tests/`. Never run `shfmt -w`.
- One issue = one PR.

w3lld1:
## Background

Related #1026

Numeric assertions covered open-ended comparisons but not a single inclusive range check. I added bounded assertions so one logical range check reports one result and one failure message.

## Changes

- I added `assert_between <min> <max> <actual>` with inclusive bounds and `assert_not_between` as its exact negation.
- I reused the fork-free fixed-point comparison path for common integers and decimals, with the existing `bc`/`awk` fallback for unsupported precision.
- I return usage errors for missing arguments, non-numeric values, and reversed bounds.
- I added coverage for boundaries, decimals, negative values, failure details, assertion counts, arity, and invalid input.
- I updated the assertion docs, generated CLI snapshot, completions, catalogue counts, and changelog.

## Verification

- `make test` — 1,789 passed; 32 skipped; 4 incomplete; 2 snapshots; no failures
- `./bashunit --parallel tests/` — 1,751 passed; 35 skipped; 4 incomplete; 2 snapshots; no failures
- focused assertion, arity, completion, and fork-budget suites — 85 passed; 1 skipped; no failures
- `make sa`
- `make lint`
- `git diff --check`

## Checklist

Fixes #1026


Chemaclass (approved):
Approved.

Solid feature. Inclusive bounds, an exact negation, usage errors for reversed and non-numeric input, and a failure message that names which bound broke. Docs, both completion files, the doc snapshot, the arity provider and the CHANGELOG are all updated.

Verified locally: full suite, `--parallel`, `make sa`, `make lint`, and no new shfmt drift against `main`.

I pushed one fix on top (fb697ece).

`assert_between "+1" "+9999999999999999999999" "+5"` leaked a raw `bc` parse error and then reported a bogus `expects min <= max` usage error. Operands wider than the fork-free fixed-point path fall through to `bc`, and `bc` cannot parse a leading `+`: it returns an empty result, which `is_le` reads as "greater than". `assert_within_delta` already strips the sign for exactly this reason, so `bashunit::math::is_le` now does it for every caller. Regression tests in `math_test.sh` and `numeric_test.sh`.

I also pinned the `_is_numeric` hardening this branch introduces. That is a user-visible fix on its own and it was not recorded: on `main`, `assert_within_delta "1.2.3" "1" "0.5"` leaks `Parse error: bad expression` into the report, and `5-3` is silently evaluated as `2`. Two regression tests plus a Fixed entry in the CHANGELOG.

PR #1047, opened 2026-08-10
test(build): raise the standalone size guard to 550 KiB

Chemaclass:
## 🤔 Background

Build Verification has been red on `main` since #1037: the built binary is 513 KiB against a 512 KiB guard. The budget landed in #990 with the binary at 485 KiB, and the features merged since (#1033-#1042) grew it past the line.

## 💡 Changes

- Re-baseline the standalone size guard to 550 KiB, ~9% headroom over the current 513 KiB build
- Document that the guard catches accidental bloat (a stray embed, a disabled comment strip) rather than capping planned growth, and how to move it again

### DIFF
--- a/tests/unit/project/build_test.sh
+++ b/tests/unit/project/build_test.sh
@@ -321,7 +321,10 @@ function test_built_binary_contains_no_source_lines() {
   assert_equals "0" "$(grep -c '^source ' "$build_dir/bashunit")"
 }
 
-function test_built_binary_stays_below_500_kib() {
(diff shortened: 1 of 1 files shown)


Chemaclass:
Superseded by #1046, which raised the guard to 544 KiB. Current `main` (217c6d45) builds to 532585 bytes, leaving ~24 KB of headroom, so the further bump to 550 KiB is not needed yet and would only blunt the growth alarm. Reopen if the artifact starts pressing against 544 KiB.

PR #1069, opened 2026-08-12
feat(assert): add JSON absence and length assertions

### LINKED ISSUE #1025: feat(assert): assert_json_key_not_exists and assert_json_length
## Problem

JSON support is three assertions (`src/assert/json.sh`): `assert_json_key_exists` (:11), `assert_json_contains` (:27), `assert_json_equals` (:49). Two everyday checks have no expression:

- **A key must be absent.** Asserting that a response does *not* leak `password`, or that an optional field was correctly omitted, has no positive form — `assert_json_key_exists` has no negative counterpart. Every other assertion family in bashunit ships its negation.
- **Collection size.** Asserting "the API returned 3 items" means `assert_equals 3 "$(echo "$json" | jq '.items | length')"`, which puts the jq invocation and its failure modes in the test instead of in the framework.

## Proposal

```bash
assert_json_key_not_exists '.user.password' "$json"
assert_json_length 3 '.items' "$json"
```

- Argument order must match the existing json assertions exactly — read `assert_json_key_exists` first and follow it, do not invent a new order.
- `assert_json_length` works on arrays and on objects (`jq length` semantics), and on strings if that falls out naturally; document what it counts.
- Both require `jq` and must go through `bashunit::assert_json::require_jq` (`src/assert/json.sh:3`) so a missing `jq` produces the same message as the existing three.
- Unparseable JSON must **fail**, not pass — this was fixed for `assert_json_equals` in #967 and the same rule applies here.

## Acceptance criteria

- [ ] `assert_json_key_not_exists` passes for an absent key and fails for a present one
- [ ] It fails (does not pass) when the JSON is unparseable (#967 rule)
- [ ] A key present with value `null` — decide, document and test whether that counts as existing
- [ ] `assert_json_length` passes for arrays and objects of the expected size
- [ ] `assert_json_length` on a path that does not exist fails with a clear message rather than comparing against empty
- [ ] A non-numeric expected length is a usage error (#984 precedent)
- [ ] Missing `jq` produces the same message as the existing json assertions
- [ ] Argument order verified against `assert_json_key_exists` (the spy family's inconsistent argument order, #984, is the anti-pattern to avoid repeating)

## Repo checklist (agent)

- **TDD**: RED → GREEN → REFACTOR. Use `/add-assertion` if helpful.
- **Bash 3.0+ only**: no `printf -v`, no `+=`, no `declare -A`, no `[[ ]]`, no `${var,,}`, no `&>>`.
- Assertions are **bare-named** (`assert_x`), helpers are `bashunit::`-namespaced. Start every assertion with `bashunit::assert::should_skip && return 0` and report through `bashunit::assertion_failed` / `bashunit::assertion_passed`, matching its neighbours in the same file.
- A missing required argument must go through `bashunit::assert::usage_error`, not compare against an empty string (#983).
- **Tests**: mirror the src layout — `tests/unit/assert/<file>_test.sh`. There is no `assert_fails`; test failure output by comparing against `print_failed_test`, as the existing tests do.
- **Docs**: add the entry to `docs/assertions.md`, then **regenerate the `bashunit doc` acceptance snapshot** (editing that file breaks it).
- Update the assertion count in `README.md` ("It ships 73 assertions…" — currently already stale at 74) and add a `CHANGELOG.md` line under `## Unreleased`.
- **Gates**: `make sa`, `make lint`, `./bashunit tests/`, `./bashunit --parallel tests/`. Never run `shfmt -w`.
- One issue = one PR.

ColumbusLabs:
## Background

Related #1025

Adds the missing positive APIs for asserting that a singular jq path is absent and that a collection or string at a path has an expected length.

## Changes

- add `assert_json_key_not_exists`, treating keys with `null` values as present and rejecting invalid JSON
- add `assert_json_length` for arrays, objects, and strings, with clear failures for missing paths and invalid input
- add Bash 3-compatible unit and arity coverage, shell completions, assertion docs, generated CLI-doc snapshot, changelog entry, and updated assertion count

## Validation

- `make sa`
- `make lint`
- `env -u NO_COLOR ./bashunit tests/ --simple` (2,094 tests; 2,071 passed, 17 skipped, 4 incomplete, 2 expected snapshots)
- `env -u NO_COLOR ./bashunit --parallel tests/ --simple` (2,050 tests; 2,029 passed, 15 skipped, 4 incomplete, 2 expected snapshots)
- `cd docs && npm ci && npm run build`

## Checklist

Closes #1025


Chemaclass (approved):
Thanks!!

JesusValeraDev (approved):

